PR #63, opened 2016-09-06
Implement a Go Spark CLI

nickbp:
For now this focuses on being a drop-in replacement for the current
CLI, where an end user shouldn't notice any difference, except that
the supported 'spark-submit' args now show up in the CLI's `--help`.

The implementation provides the same arguments that spark-submit uses,
but just converts those into a POST request that's sent to the
dispatcher. This doesn't support things like maven lookup of jar
dependencies, but those things were only applicable to client mode, or
things like YARN, anyway.

This implementation supports both the Binary CLI in 1.8+, while also
providing a simple python wrapper package for backwards compatibility
for 1.7. The stub-universe packaging will automatically fill in the
required binary CLI SHAs for us.


nickbp:
Added fixes to `jenkins-package-test.sh` (recent changes to S3_BUCKET/S3_PREFIX appeared to be causing problems?):
- test.sh was specifically wanting `DEV_` prefixes on AWS stuff
- set the `S3_PREFIX` to go within `/autodelete7d/` so that ephemeral builds aren't accumulating in S3
- the `VERSION` wasn't being set. went with using the SHA. also used this value in `S3_PREFIX` to avoid builds stepping on each other

nickbp:
retest this please

mrbrowning:
I'll wait on tests to pass, but seems fine.

nickbp:
Alright, green once more

cli/bin/build-go.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+set +x
+
+BIN_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $BIN_DIR/..
```
mrbrowning: It *should* never happen, but `$BIN_DIR` could contain a space, so let's quote it. I think it's worth being maximally defensive in Bash scripts.

nickbp: Done

PR #65, opened 2016-09-22
[SPARK-339] Add support for Kerberos args to Go CLI module

nickbp:
Tested by launching a job with phony Kerberos args against the Python CLI and then the Go CLI with the same args, then checking the resulting launched job args for differences. From that testing, assuming Python submit works, Go submit should work too..

### DIFF
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/mesosphere/dcos-commons/cli"
 	"gopkg.in/alecthomas/kingpin.v2"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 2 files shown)


nickbp:
build failure: `ERROR: Unable to find project for artifact copy: mesos-spark-integration-tests`

cli/dcos-spark/submit_builder.go
```diff
@@ -204,7 +228,7 @@ ARGLOOP:
 		}
 	}
 	if cli.Verbose {
-		fmt.Printf("Translated arguments: '%s'\n", argsEquals)
+		log.Printf("Translated arguments: '%s'\n", argsEquals)
```
nickbp: FYI this change is to move a stdout print to stderr.

nickbp:
retest this please

elezar:
@nickbp Since the JIRA ticket has been closed and this PR is pretty stale, I assume that we can close this? Please do so, and create a follow up ticket if there are any changes that should still be incorporated in our Spark package.

nickbp:
I think this was done in a separate PR. In any case it's very old. Closing.

nickbp:
Yeah it looks like related work was done via #257

PR #78, opened 2016-10-24
Rough draft documenting Python support and Spark shell.

susanxhuynh:
@sascala Here is my initial draft.

### DIFF
--- a/docs/user-docs.md
+++ b/docs/user-docs.md
@@ -18,6 +18,7 @@ DC/OS Spark includes:
 *   [Mesos Cluster Dispatcher][2]
 *   [Spark History Server][3]
 *   DC/OS Spark CLI
+*   Interactive Spark shell
 
 ## Benefits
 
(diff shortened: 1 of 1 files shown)


docs/user-docs.md
```diff
@@ -59,6 +59,10 @@ dispatcher and the history server
 
         $ dcos spark run --submit-args="--class org.apache.spark.examples.SparkPi http://downloads.mesosphere.com.s3.amazonaws.com/assets/spark/spark-examples_2.10-1.4.0-SNAPSHOT.jar 30"
 
+1.  Run a Python Spark job:
+
+        $ dcos spark run --submit-args="http://xhuynh-dev.s3.amazonaws.com/pi.py 30"
```
sascala: It seems we shouldn't be using s3 URLS (see: https://mesosphere.atlassian.net/browse/INFINITY-575?filter=-1) -- I've been meaning to check in with you or @mgummelt about what the correct links ought to be.

PR #84, opened 2016-11-04
Ss/auto update

sascala:
I've modified the spark docs to make it possible to automatically import them to the enterprise site. There are a lot of aesthetic changes that I don't think need an expert eye, but I'd appreciate a glance to make sure things like the page ordering (see the TOC) make sense.


NimaVaziri:
The restructuring makes sense. Only thing that stood out to me was that as a user I'd expect things like `Runtime Configuration Change`, Run a Spark Job`, and `Interactive Spark Shell` to be slightly higher in the TOC.

Otherwise, once the branch conflicts are resolved, LGTM.

PR #88, opened 2016-11-08
Add support for testing against strict mode clusters

nickbp:


bin/test.sh
```diff
@@ -63,12 +63,15 @@ configure_cli() {
 install_spark() {
     notify_github pending "Installing Spark"
 
-    # with universe server running, there are no longer enough CPUs to
-    # launch spark jobs if we give the dispatcher an entire CPU
-    # TODO: remove this?
-    echo '{"service": {"cpus": 0.1}}' > /tmp/spark.json
+    if [ "$SECURITY" = "strict" ]; then
```
susanxhuynh: @nickbp When do we want to test with strict mode? Is it okay that it's not run by default?

bin/test.sh
```diff
@@ -63,12 +63,15 @@ configure_cli() {
...
+        # custom configuration to enable auth stuff:
+        ${REPO_ROOT_DIR}/dcos-commons-tools/setup_permissions.sh nobody "*" # spark's default service.role
```
susanxhuynh: Where is REPO_ROOT_DIR set?

bin/test.sh
nickbp: There's a separate job in jenkins that tests a given DC/OS build against package builds. It wouldn't be against the spark project CI but it'd be getting run periodically separately.

nickbp: Done, added further down. run-tests.py would likely encompass all of configure_cli/install_spark/run_tests.

mgummelt:
LGTM

PR #94, opened 2016-11-11
Multiple hadoop profiles

mgummelt:
This makes it so the spark-release job will publish a version of Spark for each hadoop profile (2.4, 2.6, 2.7).

It also makes it so the spark-build-release job will publish a docker image for each hadoop profile.  We still only publish one DC/OS package per release, but the user can customize the DC/OS package with a different docker image via `service.docker-image`.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -50,6 +50,23 @@ Run the following command to see all configuration options:
 
     $ dcos package describe spark --config
 
+## Customize Spark Distribution
+
+DC/OS Spark does not support arbitrary Spark distributions, but
+Mesosphere does provide multiple pre-built distributions, primarily
+used to select Hadoop versions.  To use one of these distributions,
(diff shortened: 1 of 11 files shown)


docs/install.md
```diff
@@ -48,7 +48,24 @@ Then, install Spark with your custom configuration:
...
 
-    $ dcos package describe spark --config
+$ dcos package describe spark --config
+
+## Customize Spark Distribution
+
+DC/OS Spark does not support arbitrary Spark distributions, but
+Mesosphere does provide multiple pre-built distributions, primarily
+used to select Hadoop versions.  To use one of these distributions,
+first select your desired Spark distribution from here: "<link>", then
```
mgummelt: cc @sascala How do I do an internal link?  I want to link to the new spark-versions.md file in this PR.

mgummelt:
cc @susanxhuynh for review

PR #96, opened 2016-11-14
DONOTSUBMIT Test PR to exercise CI

nickbp:
Check that PR CI still works with recent CI change


nickbp:
looks like we're still good

PR #108, opened 2017-01-04
[SPARK-399] install master CLI

mgummelt:
We're testing against master DC/OS, so for consistency, we should test against the master DC/OS CLI as well.  The immediate benefit is that this should fix the logging problem we see in the tests.


mgummelt:
blocked on https://github.com/dcos/dcos-cli/pull/866

mgummelt:
cc @susanxhuynh for review

bin/jenkins.sh
```diff
@@ -68,14 +68,11 @@ function docker_version() {
...
-    mkdir dcos-cli
-    bash install.sh dcos-cli http://change.me --add-path no
-    source dcos-cli/bin/env-setup
-
-    # hack because the installer forces an old CLI version
-    pip install -U dcoscli
+    git clone git@github.com:dcos/dcos-cli.git
+
+    (cd dcos-cli && make env)
+    (cd dcos-cli/cli && make env)
```
susanxhuynh: Should this be "cd cli" since it's already in the dcos-cli directory?

bin/jenkins.sh
```diff
@@ -68,14 +68,11 @@ function docker_version() {
...
+    source dcos-cli/cli/bin/env-setup-dev
```
susanxhuynh: What's the difference between this "env-setup-dev" vs. "env-setup"?

bin/jenkins.sh
mgummelt: The parens means those commands are executed in an isolated environment.  It doesn't change the CWD for subsequent commands

bin/jenkins.sh
mgummelt: Users use to use `env-setup` when we had them install via python instead of via the native CLI.   It's a different set of deps.

susanxhuynh:
LGTM

mgummelt:
retest this please

mgummelt:
Turns out this doesn't fix the problem, because shakedown uses its own (non-master) version of the CLI.

PR #121, opened 2017-02-22
Fix the shakedown tests for strict mode

susanxhuynh:
- Temporarily using a stub-universe for HDFS until strict mode support is released, just to get Framework Test to pass. Should be reverted when a new version of HDFS is released.
- Specify the principal and secret when install HDFS in strict mode.
- Specify security configs for Spark driver in strict mode.

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -64,7 +64,7 @@ setup_permissions() {
     if [ "$SECURITY" = "strict" ]; then
         # custom configuration to enable auth stuff:
         ${COMMONS_TOOLS_DIR}/setup_permissions.sh nobody "*" # spark's default service.role
-        ${COMMONS_TOOLS_DIR}/setup_permissions.sh nobody hdfs-role
+        ${COMMONS_TOOLS_DIR}/setup_permissions.sh root hdfs-role
     fi
 }
 
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -77,7 +77,7 @@ run_tests() {
     fi
     source env/bin/activate
     pip install -r requirements.txt
-    py.test test.py
+    py.test -s test.py
     if [ $? -ne 0 ]; then
         notify_github failure "Tests failed"
         exit 1
--- a/tests/test.py
+++ b/tests/test.py
@@ -62,7 +62,10 @@ def _require_package(pkg_name):
             'https://infinity-artifacts.s3.amazonaws.com/master-latest-nightly/hdfs/stub-universe-hdfs.zip',
             0)
         time.sleep(90)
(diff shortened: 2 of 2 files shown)


susanxhuynh:
I removed the hdfs master stub repo. The hdfs dependence is currently in the test setup_module() method. The test currently fails during the HDFS install, before any tests are run. Are you suggesting moving that into the Teragen test, and disabling that test temporarily?  I could do that, but since the new hdfs will be released anytime now, I was thinking just wait for it to be released and not disable any tests.

tests/test.py
```diff
@@ -56,7 +57,10 @@ def _require_package(pkg_name):
     pkg_manager = dcos.package.get_package_manager()
     installed_pkgs = dcos.package.installed_packages(pkg_manager, None, None, False)
     if not any(pkg['name'] == pkg_name for pkg in installed_pkgs):
-        shakedown.install_package(pkg_name, wait_for_completion=True)
+        options = {}
```
mgummelt: This isn't guaranteed to be correct for an arbitrary package, so let's not include these options in this function.  Pass them as an argument to the function through `_require_package('hdfs')`.

Also, let's create the options functionally:
```
if (...):
  options = ...
else:
  options = ...
```

tests/test.py
```diff
@@ -154,6 +158,11 @@ def _run_tests(app_url, app_args, expected_output, args={}, config={}):
 
 
 def _submit_job(app_url, app_args, args={}, config={}):
+    if os.environ.get('SECURITY') == 'strict':
```
mgummelt: let's factor this out into an `_is_strict()` function, and replace all 3 occurrences

tests/test.py
susanxhuynh: Done

tests/test.py
susanxhuynh: Done

bin/test.sh
```diff
@@ -64,7 +64,7 @@ setup_permissions() {
     if [ "$SECURITY" = "strict" ]; then
         # custom configuration to enable auth stuff:
         ${COMMONS_TOOLS_DIR}/setup_permissions.sh nobody "*" # spark's default service.role
-        ${COMMONS_TOOLS_DIR}/setup_permissions.sh nobody hdfs-role
+        ${COMMONS_TOOLS_DIR}/setup_permissions.sh root hdfs-role
```
mgummelt: @mohitsoni, @susanxhuynh says you instructed her to setup these permissions for root in hdfs in strict mode.  Why?  Aren't root users forbidden in strict mode?

PR #122, opened 2017-03-03
[SPARK-424] Run the Spark Dispatcher as root in strict mode

mgummelt:
This PR runs the Spark Dispatcher as root to avoid the Mesos bugs encountered in strict mode, and it launches the test job with the auth token so that it can fetch state.json.

blocked on:
- https://github.com/mesosphere/dcos-commons/pull/645
- https://github.com/typesafehub/mesos-spark-integration-tests/pull/94


bin/test.sh
```diff
@@ -21,76 +21,51 @@ check_env() {
...
     py.test -s test.py
-    if [ $? -ne 0 ]; then
-        notify_github failure "Tests failed"
-        exit 1
-    fi
     popd
 }
 
 check_env
-fetch_commons_tools
```
susanxhuynh: Where do you fetch the dcos-commons tools?

bin/test.sh
```diff
@@ -21,76 +21,51 @@ check_env() {
...
     fi
 }
 
-setup_permissions() {
+initialize_service_account() {
     if [ "$SECURITY" = "strict" ]; then
-        # custom configuration to enable auth stuff:
-        ${COMMONS_TOOLS_DIR}/setup_permissions.sh nobody "*" # spark's default service.role
-        ${COMMONS_TOOLS_DIR}/setup_permissions.sh root hdfs-role
+        ${COMMONS_DIR}/tools/create_service_account.sh --strict
```
susanxhuynh: This is already called by launch_ccm_cluster.sh

tests/test.py
```diff
@@ -220,6 +182,100 @@ def _get_content_type(basename):
...
+        LOGGER.info("Installing package {}".format(pkg_name))
+        shakedown.install_package(
+            pkg_name,
+            options_json=options,
+            wait_for_completion=True)
+
+
+def _is_strict():
+    return os.environ.get('SECURITY') == 'strict'
+
```
susanxhuynh: This function is already defined above.

mgummelt: good catch.  removed.

bin/test.sh
mgummelt: I added the COMMONS_DIR environment variable, and I fetch the dcos-commons dir in the Jenkins PR project: https://jenkins.mesosphere.com/service/jenkins/view/Infinity/job/spark/job/spark-build/job/spark-build-PR/

I'll add it to the other builds as well once this passes.

The github project is the source of truth, so we should use it.  I see no point in the dcos-commons-tools tarball.

susanxhuynh: OK

package/config.json
```diff
@@ -6,7 +6,7 @@
             "type": "object",
             "properties": {
                 "name": {
-                    "default": "spark",
+                    "default": "spark-dispatcher",
```
mgummelt: I need to look into this before merging.

PR #130, opened 2017-03-21
[SPARK-437] fix strict mode in python tests

mgummelt:
@susanxhuynh are the other frameworks running strict mode in PR tests?  I'm wondering if we could have caught this sooner.


susanxhuynh:
@mgummelt None of the frameworks is running strict mode in PR tests, unfortunately.

PR #138, opened 2017-04-18
[SPARK-449] Enable check_env to work.

jsrodman:
really unfortunate bourne shell wart:

```bash
if [ condition ]; then 
   x
fi
```

Will run x if condition is true; however:

```bash
if [ condition]; then
    x
fi
```

will simply never run x because the `[` command considers this a syntax error, and those are false.


mgummelt:
Your branch is out of date.  That space exists on `master`.

jsrodman:
Done, then.

PR #139, opened 2017-04-19
[SPARK-448] Don't run the janky mesos_master configuration; not needed and fragile.

jsrodman:
don't invoke the fragile modify_master.py

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -28,7 +28,7 @@ start_cluster() {
         echo "Using existing cluster: $DCOS_URL"
     else
         echo "Launching new cluster"
-        CCM_JSON=$(CCM_AGENTS=5 ${COMMONS_DIR}/tools/launch_ccm_cluster.py)
+        CCM_JSON=$(CCM_AGENTS=5 ${COMMONS_DIR}/tools/launch_ccm_cluster.py --configure=nomaster)
         DCOS_URL=$(echo "${CCM_JSON}" | jq .url)
         if [ $? -ne 0 -o "$DCOS_URL" = "http://" ]; then
             exit 1


mgummelt:
So the workaround is just to not run the code that requires a virtualenv?  That doesn't seem ideal, since now we can only run launch_ccm_cluster.py in certain ways depending on the environment (whether or not pip is installed).  What would be the downside of just doing virtualenvs the standard way rather than using the `venv` library?

jsrodman:
It's a workaround.  By definition it's not ideal.

This is the third time i've had to work around modify_master, and I don't have access to reproduce the problem right now (question to quality outstanding) so figuring out how to do it properly today seems less important than getting spark & cassandra going at all.

As of python3.3, venv is pretty much the standard way.  If the trouble ends up being specific to venv, I'm not opposed to using the purportedly legacy virtualenv, but before casting about blindly, I wanted to get things working.

There's also no reason we can't use modify_master.py after launch_ccm_cluster completes, with a virtualenv created via any particular method.  I'd like to get the cluster configuration out of launch_ccm_cluster, since explicitly configuring the cluster makes way more sense to me.  For starters, if we have standard configurations our tests end up expecting for clusters, we should be able to invoke the cluster configuration on clusters allocated via other means.

mgummelt:
Oh I didn't realize `venv` was standard now.  I was going to propose we use `python -m venv`, but I guess that would have the same problem.  Did you find the root cause of the Spark problem?  I see in the JIRA you mentioned it arises from Spark explicitly downloading pip.  Can you link me to that code?

jsrodman:
Is this still a packaging problem?
```
++ pwd
+ SCALA_TEST_JAR_PATH=/jenkins/workspace/spark/spark-build/spark-build-PR/spark-build/tests/jobs/scala/target/scala-2.11/dcos-spark-scala-tests-assembly-0.1-SNAPSHOT.jar
+ py.test -s test.py
============================= test session starts ==============================
platform linux -- Python 3.5.2, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: /jenkins/workspace/spark/spark-build/spark-build-PR/spark-build/tests, inifile:
plugins: timeout-1.2.0
collected 7 items

test.py INFO:test:Ensuring HDFS is installed.
INFO:test:Installing package hdfs
EEEsEEE

==================================== ERRORS ====================================
__________________________ ERROR at setup of test_jar __________________________

module = <module 'test' (<_pytest.assertion.rewrite.AssertionRewritingHook object at 0x7f42d8504f98>)>

    def setup_module(module):
        if _hdfs_enabled():
>           _require_hdfs()
...
>           raise DCOSException(error_msg)
E           dcos.errors.DCOSException: Package [hdfs] not found
```
Or is it a cli problem?

jsrodman:
retest this please

mgummelt:
It's because HDFS has been effectively removed from the universe: https://github.com/mesosphere/universe/tree/version-3.x/repo/packages/H/hdfs

I suppose we'll need to update the tests to install "beta-hdfs"

jsrodman:
@mgummelt : re who's doing the pip setup; i misunderstood and it seems to be

spark-build/bin/jenkins.sh install_cli which fetches and executes install.sh from downloads.mesosphere.io, which creates a virtualenv with the default command and python which on debian are python2.

mgummelt:
@jsrodman Ah, well we should be using the binary CLI anyway.  Can you update this PR to install the binary CLI rather than the python CLI?  Then we shouldn't need the launch_ccm_cluster.py workaround.

jsrodman:
launch_ccm_cluster.py (or more accurately configure_test_cluster.py) already fetches the binary CLI as necessary after allocation the cluster, so that it knows the correct version of the cli to download by fetching the cluster version first.  I think this is the right way to do it.  Meanwhile over in INFINITY-1370 I'm arranging to download the binary CLI once and only once.  Thus changing the download method here would either introduce logic duplication or be at risk of further breakage with the fragile relationship between the dcos-commons.zip and the scripts here.  

For these reasons, I don't believe it makes sense to increase the scope.

mgummelt:
> Meanwhile over in INFINITY-1370 I'm arranging to download the binary CLI once and only once.

Where is that download going to occur?  `test.sh`?

jsrodman:
Yeah, when invoking test.sh, test.sh will arrange for the cli to be fetched and inserted on the path, and when invoking test.py, the same.  For now I'm inserting an env var in test.h saying essentially 'this is the working dir for this test run" and other components will simply verify it's there (or do whatever they currently do to fetch a cli).

You of course can review those changes when they are ready for review.

mgummelt:
That won't help spark, unfortunately, since it doesn't use test.sh.  I have
a pr out to run the tests via run_tests.sh, which should allow me to remove
the current cli install from the Jenkins test. I'll see if that fixes it.

On Apr 19, 2017 7:46 PM, "Joshua Rodman" <notifications@github.com> wrote:

Yeah, when invoking test.sh, test.sh will arrange for the cli to be fetched
and inserted on the path, and when invoking test.py, the same. For now I'm
inserting an env var in test.h saying essentially 'this is the working dir
for this test run" and other components will simply verify it's there (or
do whatever they currently do to fetch a cli).

You of course can review those changes when they are ready for review.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mesosphere/spark-build/pull/139#issuecomment-295557574>,
or mute the thread

jsrodman:
I was presuming you had your own reasons to fetch the cli.  

One of the goals is to move away from run_test.py downloading the cli 7 times for 7 frameworks, not that this is relevant for `spark-build`.  I interpreted your suggestion for downloading the binary cli as replacing the existing function in your existing scripts to use tools/cli_install.py.  That's already possible; but run_test.py will just download it again as-is, so maybe it's pointless for this repo.

mgummelt:
Spark fetches it so it can configure the cluster, but run_tests.py also
does that.

mgummelt:
Closing this, since SPARK-451 should resolve this problem.

PR #140, opened 2017-04-20
[SPARK-451] Use run-tests.py to run tests

mgummelt:
`launch_ccm_cluster.py` introduced a breaking change where it fails w/ certain versions of pip are already installed, which they are in the Spark builds because we've been installing the old python version of the CLI.  @jsrodman proposed a fix (https://github.com/mesosphere/spark-build/pull/139) which would make Spark avoid the `venv` code affected by this change, but I'd rather not create yet another one-off for how Spark runs things.  The fix in this PR removes the CLI install entirely, and instead runs the tests with `run_tests.py`, which itself does a CLI install.

cc @ArtRand @jsrodman


tests/test_spark.py
```diff
@@ -187,17 +198,18 @@ def _get_hdfs_options():
     if _is_strict():
         options = {'service': {'principal': 'service-acct', 'secret_name': 'secret'}}
     else:
-        options = {}
+        options = {"service": {}}
+
+    options["service"]["beta-optin"] = True
```
mgummelt: The beta-hdfs changes in this file are just to get the tests working with the latest breaking hdfs change.

mgummelt:
retest this please

PR #145, opened 2017-04-28
Added terasort test for soak cluster

susanxhuynh:
This test runs teragen, terasort, and teravalidate. Because the soak cluster is long-running, we have to delete the output files from HDFS after each test run. This is done by the hdfs-delete-terasort-files job.
I added a "DO_TEARDOWN" env to indicate that packages are not to be uninstalled in soak cluster testing.


tests/test_spark.py
```diff
@@ -169,6 +216,17 @@ def _require_spark():
     _wait_for_spark()
 
 
+def _require_spark_cli():
```
mgummelt: Is this needed for the soak tests?  I'd rather have the test runner code be responsible for setting up the environment (incuding the CLI), rather than the tests themselves.

tests/test_spark.py
```diff
@@ -35,19 +36,23 @@ def _init_logging():
 HDFS_PACKAGE_NAME='beta-hdfs'
 HDFS_SERVICE_NAME='hdfs'
 SPARK_PACKAGE_NAME='spark'
+TERASORT_JAR='https://downloads.mesosphere.io/spark/examples/spark-terasort-1.0-jar-with-dependencies_2.11.jar'
+TERASORT_MAX_EXECUTOR_CORES=6
```
mgummelt: Can we call this `TERASORT_MAX_CORES` instead?  It reads as if it's the max number of cores a single executor will have, but that's not the effect.

tests/test_spark.py
```diff
@@ -66,11 +71,53 @@ def test_jar():
...
+               "{} hdfs:///terasort_in".format(input_size),
+               "Number of records written",
+               ["--class", "com.github.ehiggs.spark.terasort.TeraGen",
+                "--conf", "spark.cores.max={}".format(TERASORT_MAX_EXECUTOR_CORES)])
+
+
+@pytest.mark.soak
+def test_terasort():
+    if _hdfs_enabled():
+        _delete_hdfs_terasort_files()
```
mgummelt: nice organization!

tests/dcosjob.py
```diff
@@ -0,0 +1,37 @@
+import os
+import shakedown
+
+
+def add_job(job_name):
+    jobs_folder = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), 'jobs', 'json'
+    )
+    _run_cli('job add {}'.format(
```
mgummelt: Is there no `dcos` or `shakedown` function for this?

tests/dcosjob.py
```diff
@@ -0,0 +1,37 @@
+import os
```
mgummelt: Ideally, this file wouldn't exist, and we would be calling into `shakedown` or `dcos` for all these functions.

tests/test_spark.py
```diff
@@ -66,11 +71,53 @@ def test_jar():
...
     if _hdfs_enabled():
-        jar_url = "https://downloads.mesosphere.io/spark/examples/spark-terasort-1.0-jar-with-dependencies_2.11.jar"
-        _run_tests(jar_url,
-                   "1g hdfs:///terasort_in",
-                   "Number of records written",
-                   ["--class", "com.github.ehiggs.spark.terasort.TeraGen"])
+        _run_teragen()
+
+
+def _run_teragen():
```
mgummelt: I think all the other private functions are at the bottom of this file.  Can we maintain that organization?

Actually, I think ideally, the terasort test would be in a separate `test_soak.py` file.  Then the tests would be more organized by their application.

susanxhuynh: I'll put the terasort test in a separate file.

tests/test_spark.py
mgummelt: Oh, I see.  This function is necessary because `_require_spark` will not install the CLI if the marathon app for the package is already installed then, right?  If so, can you just move this functionality to _require_spark()?  I'd rather have that function just be fully responsible for installing the entire package, including the CLI.

tests/dcosjob.py
mgummelt: What's the underlying function that the dcos cli is using to check that a job is finished?  If that exists in the `dcos` package, let's use that.  Otherwise, CLI is fine.

susanxhuynh:
@mgummelt Addressed your review comments:
- I reorganized the Terasort code into a separate file, test_soak.py.
- Added a new "utils.py" file for common utility functions.

tests/test_soak.py
```diff
@@ -0,0 +1,109 @@
...
+    metronome_client.add_job(job)
+
+
+def _run_job_and_wait(metronome_client, job_name, timeout_seconds):
+    metronome_client.run_job(job_name)
+
+    shakedown.wait_for(
+        lambda: (
+            'Successful runs: 1' in
+            _run_cli('job history {}'.format(job_name))
```
susanxhuynh: Checking a job's completion status is messy in the CLI; it hits a `jobs` endpoint and parses the results: https://github.com/dcos/dcos-cli/blob/0.4.16/cli/dcoscli/job/main.py#L266 I will use the CLI for now.

tests/test_soak.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.mark.soak
+def test_terasort():
+    if utils.hdfs_enabled():
+        _delete_hdfs_terasort_files()
+        _run_teragen()
+        _run_terasort()
+        _run_teravalidate()
+
+
+def _require_spark_cli():
```
susanxhuynh: Do you still want to combine this method with require_spark(), which is in another file?

mgummelt: yes please

tests/test_soak.py
mgummelt: :+1:

PR #150, opened 2017-06-05
Check if the hdfs_delete_terasort_files job exists before adding it

susanxhuynh:
Metronome throws an exception if you try to add an existing job.
To avoid this, I added a check of whether the hdfs_delete_terasort_files job exists before adding it.


mgummelt:
retest this please

PR #156, opened 2017-06-15
[DOCS-1985] [WIP] Spark doc fixup

joel-hamill:
https://jira.mesosphere.com/browse/DOCS-1985


docs/custom-docker.md
```diff
@@ -19,9 +19,9 @@ You can customize the Docker image in which Spark runs by extending the standard
 
 1. Then, build an image from your Dockerfile.
 
-        $ docker build -t username/image:tag .
-        $ docker push username/image:tag
+        docker build -t username/image:tag .
```
mgummelt: Why remove the shell prompt?

docs/hdfs.md
```diff
@@ -5,16 +5,19 @@ menu_order: 20
...
+You can configure Spark for a specific HDFS cluster.
 
-    {
-      "hdfs": {
-        "config-url": "http://mydomain.com/hdfs-config"
-      }
-    }
+To configure `hdfs.config-url` to be a URL that serves your `hdfs-site.xml` and `core-site.xml`, use this example where `http://mydomain.com/hdfs-config/hdfs-site.xml` and `http://mydomain.com/hdfs-config/core-site.xml` are valid URLs:
 
+```json
```
mgummelt: I think @sascala told me that these language indentifiers are Github specific, and don't work when we translate these to the service docs webpage.  Is that still the case?

docs/index.md
```diff
@@ -5,49 +5,4 @@ feature_maturity: stable
...
-
- [1]: http://spark.apache.org/documentation.html
- [2]: http://spark.apache.org/docs/latest/running-on-mesos.html#cluster-mode
- [3]: http://spark.apache.org/docs/latest/monitoring.html#viewing-after-the-fact
- [4]: https://docs.mesosphere.com/service-docs/hdfs/
- [5]: https://docs.mesosphere.com/service-docs/kafka/
- [6]: https://zeppelin.incubator.apache.org/
- [17]: https://github.com/mesosphere/spark
- [18]: https://github.com/mesosphere/spark-build
+Welcome to the documentation for the DC/OS Apache Spark version 1.0.9-2.1.0-1. For more information about new and changed features, see the [release notes](https://github.com/mesosphere/spark-build/releases/tag/1.0.9-2.1.0-1).
```
mgummelt: We can't hardcode the version in the docs.  It changes.  This one is already out of date.

docs/index.md
```diff
@@ -5,49 +5,4 @@ feature_maturity: stable
 enterprise: 'no'
 ---
 
-Apache Spark is a fast and general-purpose cluster computing system for big data. It provides high-level APIs in Scala, Java, Python, and R, and an optimized engine that supports general computation graphs for data analysis. It also supports a rich set of higher-level tools including Spark SQL for SQL and DataFrames, MLlib for machine learning, GraphX for graph processing, and Spark Streaming for stream processing. For more information, see the [Apache Spark documentation][1].
```
mgummelt: Why remove this?

docs/install.md
```diff
@@ -5,78 +5,99 @@ feature_maturity: stable
...
+Spark is available in the Universe and can be installed by using either the GUI or the DC/OS CLI.
 
-##  <a name="install-enterprise"></a>Prerequisites
+**Prerequisites:**
 
-- Depending on your security mode in Enterprise DC/OS, you may [need to provision a service account](https://docs.mesosphere.com/service-docs/spark/spark-auth/) before installing Spark. Only someone with `superuser` permission can create the service account.
-	- `strict` [security mode](https://docs.mesosphere.com/1.9/installing/custom/configuration-parameters/#security) requires a service account.  
-	- `permissive` security mode a service account is optional.
-	- `disabled` security mode does not require a service account.
+- [DC/OS and DC/OS CLI installed](https://docs.mesosphere.com/1.9/installing/).
```
mgummelt: We probably shouldn't hardcode the DC/OS version in the URL

docs/install.md
```diff
@@ -5,78 +5,99 @@ feature_maturity: stable
 enterprise: 'no'
 ---
 
-Spark is available in the Universe and can be installed by using either the web interface or the DC/OS CLI.
+Spark is available in the Universe and can be installed by using either the GUI or the DC/OS CLI.
 
-##  <a name="install-enterprise"></a>Prerequisites
```
mgummelt: What is this anchor for?

docs/hdfs.md
joel-hamill: some do, some don't. `json` does, `c++` does not.

docs/index.md
joel-hamill: isn't that what we currently have: https://docs.mesosphere.com/service-docs/spark/v1.0.9-2.1.0-1/ ?

docs/index.md
joel-hamill: duplicate info, consolidated in parent: https://github.com/mesosphere/dcos-docs-enterprise/blob/4a9345aa626f7769df078a659d84a4f31c8628a5/service-docs/spark/index.md

docs/install.md
joel-hamill: this is the only way to currently have the doc links work both in the repo, and after imported, on docs.mesosphere.com.

docs/install.md
joel-hamill: it was not referenced anywhere: https://github.com/mesosphere/spark-build/search?utf8=%E2%9C%93&q=%22install-enterprise%22&type= or https://github.com/mesosphere/universe/search?utf8=%E2%9C%93&q=%22install-enterprise%22&type=

docs/index.md
mgummelt: I beleive @sascala copies the docs for that version into the docs under that URL.  If we hardcode the version in the docs themselves, then that version string will show up in the docs for *all versions*, which is bad.

docs/index.md
mgummelt: Can you instead modify your/suzanne's docs copying script to copy this source into that location?  I think all the documentation source should live with the project, otherwise it's difficult to change as the product changes.

docs/install.md
mgummelt: Can't you use the "latest" URL?

docs/install.md
mgummelt: What do you mean?  Why is it not being a referenced a reason to add it?  Shouldn't we only add it if it *is* referenced?

joel-hamill: @mgummelt line 10 is being removed (pink highlighting). are we talking about the same line?

docs/install.md
joel-hamill: yeah, we could simply link to `https://docs.mesosphere.com/installing/`. however, since these service versions only support specific versions of DC/OS, latest won't always be correct as we iterate (e.g. 2 months from now when 1.10 is GA, this version of the spark guide should still link specifically to the 1.9 doc, correct?).

docs/install.md
mgummelt: Oh jeez, I misread this.  nvm.

docs/index.md
joel-hamill: makes sense, could we link to the high level `https://github.com/mesosphere/spark-build/releases/` page?

docs/index.md
joel-hamill: i can move back in to this project. for the dcos-docs-enterprise content i can add a simple landing page header, similar to `Welcome to the documentation for Apache DC/OS Spark. Choose a version at the left to get started!` - and leave the rest of the content in this repo.

docs/custom-docker.md
sascala: you can cut-and-paste commands more easily without the shell prompts

docs/install.md
```diff
@@ -5,78 +5,99 @@ feature_maturity: stable
...
+**Prerequisites:**
 
-- Depending on your security mode in Enterprise DC/OS, you may [need to provision a service account](https://docs.mesosphere.com/service-docs/spark/spark-auth/) before installing Spark. Only someone with `superuser` permission can create the service account.
-	- `strict` [security mode](https://docs.mesosphere.com/1.9/installing/custom/configuration-parameters/#security) requires a service account.  
-	- `permissive` security mode a service account is optional.
-	- `disabled` security mode does not require a service account.
+- [DC/OS and DC/OS CLI installed](https://docs.mesosphere.com/1.9/installing/).
+- Depending on your [security mode](https://docs.mesosphere.com/1.9/overview/security/security-modes/), Kafka requires a service authentication for access to DC/OS. For more information, see [Configuring DC/OS Access for Kafka](https://docs.mesosphere.com/service-docs/kafka/kafka-auth/).
```
sascala: "service authentication" token?

docs/quickstart.md
```diff
@@ -0,0 +1,154 @@
+---
+post_title: Spark Quickstart
+menu_order: 10
+feature_maturity: stable
+enterprise: 'no'
+---
+
+This tutorial will get you up and running in minutes with Spark. You will install the DC/OS Spark.
```
sascala: "the DC/OS Spark" service?

docs/quickstart.md
```diff
@@ -0,0 +1,154 @@
...
+
+**Prerequisites:**
+
+-  [DC/OS and DC/OS CLI installed](https://docs.mesosphere.com/1.9/installing/) with a minimum of three agent nodes with eight GB of memory and ten GB of disk available on each agent.
+-  Depending on your [security mode](https://docs.mesosphere.com/1.9/overview/security/security-modes/), Spark requires a service authentication for access to DC/OS. For more information, see [Configuring DC/OS Access for Spark](https://docs.mesosphere.com/service-docs/spark/spark-auth/).
```
sascala: "a service authentication" token?

docs/install.md
mgummelt: Ah, yea, I suppose so.

docs/install.md
```diff
@@ -5,78 +5,99 @@ feature_maturity: stable
 enterprise: 'no'
 ---
 
-Spark is available in the Universe and can be installed by using either the web interface or the DC/OS CLI.
+Spark is available in the Universe and can be installed by using either the GUI or the DC/OS CLI.
```
mgummelt: I think we refer to the web app as the "UI" in most places, not the "GUI".

docs/install.md
mgummelt: Also, this refers to "Kafka", not "Spark".

docs/quickstart.md
mgummelt: DC/OS Apache Spark, actually (Apache gets upset when we don't use their name)

docs/quickstart.md
```diff
@@ -0,0 +1,154 @@
...
+       ![Verify spark installation](/img/spark-gui-install.png)
+       
+    -  Type `dcos spark` to view the Spark CLI options.
+
+1.  Run the sample Java Spark Pi job for DC/OS. This runs a Spark job which calculates the value of Pi. You can view the example source [here](https://downloads.mesosphere.com/spark/assets/spark-examples_2.10-1.4.0-SNAPSHOT.jar). 
+
+    1.  Run this command: 
+
+        ```bash
+        dcos spark run --submit-args="--class org.apache.spark.examples.SparkPi https://downloads.mesosphere.com/spark/assets/spark-examples_2.10-1.4.0-SNAPSHOT.jar 30"
```
mgummelt: This JAR is old (2.10) and won't work against the latest Spark version.  Did you test this?

mgummelt: https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar should work

docs/quickstart.md
```diff
@@ -0,0 +1,154 @@
...
+    ```
+   
+    **Tips:** 
+    
+    -  You can view the status of your Spark installation from the DC/OS GUI **Services** tab.
+       
+       ![Verify spark installation](/img/spark-gui-install.png)
+       
+    -  Type `dcos spark` to view the Spark CLI options.
+
```
mgummelt: "Run the sample SparkPi jar".

It's not Java.  It's scala.  But the source language is irrelevant.

docs/quickstart.md
```diff
@@ -0,0 +1,154 @@
...
+        
+        Your output should resemble:
+        
+        ```bash
+        Registered docker executor on 10.0.2.199
+        Starting task driver-20170616212957-0001
+        Pi is roughly 3.141853333333333
+        ```
+
+1.  Run a Python Spark Pi job. This runs a Python Spark job which calculates the value of Pi. You can view the example source [here](https://downloads.mesosphere.com/spark/examples/pi.py).
```
mgummelt: s/Spark Pi/SparkPi

docs/quickstart.md
```diff
@@ -0,0 +1,154 @@
...
+        
+        Your output should resemble:
+        
+        ```bash
+        Registered docker executor on 10.0.2.199
+        Starting task driver-20170616213917-0002
+        Pi is roughly 3.142211
+        ```
+
+1.  Run the [SparkDataFrame](https://spark.apache.org/docs/latest/sparkr.html) example job. You can view the example source [here](https://downloads.mesosphere.com/spark/examples/dataframe.R).
```
mgummelt: "Run an R job."

docs/security.md
```diff
@@ -3,96 +3,12 @@ post_title: Security
...
-
-    <td>
-      Set to true to enable SSL on Mesos communication (default: false).
-    </td>
-  </tr>
-</table>
-
-
-## Authentication
+This topic describes how to configure security for Spark.
```
mgummelt: "configure security" is too broad.  "Configure Spark with a DC/OS service account".

docs/security.md
```diff
@@ -3,96 +3,12 @@ post_title: Security
 menu_order: 40
 enterprise: 'no'
 ---
-# Mesos Security
 
-## SSL
-
-<table class="table">
```
mgummelt: Why remove the contents of this file?  Where is the source for https://docs.mesosphere.com/service-docs/spark/spark-auth/?

I wrote this file because I think Spark-specificy security docs should live with the Spark project.

corbinpacheco:
@joel-hamill These are some thoughts that I had shared on Slack and posting here.

there are a couple basic things at the top that I think would be confusing to a nube

for example, i would separate installing the platform from installing the CLI

as a user I might have a dcos cluster already and just need to install the CLI

i have no idea what the security stuff means as a user (strict/permissive) so I would probably omit that or just point to the other doc you have for installing on a secure cluster
as a user, just 4 things I care about.

1.  Install spark service
2.  How to run a job
3.  Get status of the job
4.  get job logs

#3 is a bit light right now, so interested in what @mgummelt 
 has to say as the best way to get job status

Since we can't submit jobs in the DC/OS UI, I think we should provide the details needed to interact with the spark service entirely from the CLI.  We might show how to open the spark service UI as an option to get job status.

corbinpacheco:
For the comments I left above, I thought something like this might be useful:

Prerequisites:
A [DC/OS cluster installed](https://docs.mesosphere.com/1.9/installing/) with a minimum of three agent nodes with eight GB of memory and ten GB of disk available on each agent. 
The [DC/OS CLI](https://docs.mesosphere.com/1.9/cli/install/) installed on your local machine
Depending on your security mode in Enterprise DC/OS, you may need to provision a service account before installing Spark. Only someone with superuser permission can create the service account.
strict security mode requires a service account.
permissive security mode a service account is optional.
disabled security mode does not require a service account.
 
For more information see the following topics:
Security Modes
Provisioning Service Accounts
Deploying spark with a service account
 
This tutorial covers how to install spark with the default settings (without a service account)

docs/quickstart.md
joel-hamill: @mgummelt should this be changed globally (i.e. no instances of simply `DC/OS Spark`)?

docs/install.md
joel-hamill: For DC/OS docs, it should be `GUI`. This has not been globally fixed, yet, but is the goal.

docs/quickstart.md
joel-hamill: @mgummelt i tested against 1.9. is `https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar` valid for 1.9?

docs/quickstart.md
joel-hamill: @mgummelt `job` or `jar`?

docs/security.md
joel-hamill: @mgummelt this particular topic merely points to the Spark service authentication information, and then discusses SSL. how about `authentication` instead of `security`?

docs/security.md
joel-hamill: @mgummelt  spark-auth does "live" with this content - once it's finally rendered at docs.mesosphere.com/service-docs/spark/. do you think we should embed it directly in the service guide repo?

docs/quickstart.md
sascala: I believe we use "Apache" at first mention and then not thereafter

docs/quickstart.md
mgummelt: The DC/OS version is irrelevant.  It's the Spark version and the Jar version that must be compatible.  Any Spark version >= 2.0 will not work with a 2.10 jar

docs/quickstart.md
mgummelt: jar

docs/security.md
mgummelt: Yes, I think all docs should live with the source code for the project it's documenting.  Otherwise, the docs and product get out of sync very quickly.

docs/security.md
mgummelt: Every `authentication` is too broad.  Spark has its own internal authentication as well.  This document is only about service accounts.

mgummelt:
@corbinpacheco Can you make your comments against the text that you're referencing?  It's hard to understand exactly what you're referring to.

joel-hamill:
@corbinpacheco 

- `i would separate installing the platform from installing the CLI`: added a tip to install and quickstart
- `security stuff means as a user (strict/permissive) so I would probably omit `: depending on your security/permission, you may not be able to do any of this. i think it should remain prereq.
- `Since we can't submit jobs in the DC/OS UI, I think we should provide the details needed to interact with the spark service entirely from the CLI. `: this is included as "Next Steps".

PR #172, opened 2017-08-24
[SPARK-506] green-all-tests

ArtRand:
Refactors secrets and supervise tests to be more stable. Runs against the latest Spark distribution and checks this new functionality.


tests/test_spark.py
```diff
@@ -77,25 +77,54 @@ def test_teragen():
...
+    LOGGER.info("Job has registered")
+    shakedown.wait_for(lambda: has_running_executors(),
+                       ignore_exceptions=False,
+                       timeout_seconds=600)
+    LOGGER.info("Job has running executors")
+
+    host = shakedown.get_service("HdfsWordCount").dict()["hostname"]
+    id = shakedown.get_service("HdfsWordCount").dict()["id"]
+    driver_regex = "spark.mesos.driver.frameworkId={}".format(id)
+    shakedown.kill_process_on_host(hostname=host, pattern=driver_regex)
```
susanxhuynh: I don't understand how this pattern matching works. The pattern should match the name of the process, but I doubt that the process name has 'spark.mesos.driver.frameworkId' in it?

tests/test_spark.py
```diff
@@ -238,10 +267,12 @@ def test_marathon_group():
     #shakedown.uninstall_package_and_wait(SPARK_PACKAGE_NAME, app_id)
 
 
-@pytest.mark.skip(reason="Skip until on newer CLI so we can use dcos security ...")
+#@pytest.mark.skip(reason="Skip until on newer CLI so we can use dcos security ...")
```
susanxhuynh: Maybe change this into a `TODO:`

tests/test_spark.py
susanxhuynh: Oh, is it specified by the dispatcher?

ArtRand: Yes. I added the additional `spark.mesos.driver.frameworkId` so that we don't kill some of the infrastructure processes, there are some logging processes that have the driver string and without the additional pattern those get killed also.

tests/test_spark.py
ArtRand: Ah, no that can just go now.

PR #177, opened 2017-08-28
[SPARK-516]Use bootstrap for spark local ip

ArtRand:


ArtRand:
retest this please

universe/marathon.json.mustache
```diff
@@ -7,7 +7,7 @@
 {{#security.kerberos.krb5conf}}
         "SPARK_MESOS_KRB5_CONF_BASE64": "{{security.kerberos.krb5conf}}",
 {{/security.kerberos.krb5conf}}
-
+        "NO_BOOTSTRAP": "true",
```
susanxhuynh: If NO_BOOTSTRAP is set, then won't the `if [-z $NO_BOOTSTRAP]` above evaluate to false and skip the whole bootstrap lookup?

ArtRand: that's what `if [ -n $NO_BOOTSTRAP]` would do `-z` returns `true` when the string length is zero or unset. I agree it's a double negative, but we don't want/need to run bootstrap for the dispatcher as there is a good amount of logic in `init.sh` and `runit/run` that takes care of it. ALSO, and perhaps more importantly, it's not broken there.

susanxhuynh: Ahh, I get it now.

PR #178, opened 2017-08-30
[SPARK-514][SPARK-512] properties file test docker bug

ArtRand:
Changes `test_secrets` to use a file containing the spark properties to set. This file is passed using the `--properties-file` CLI flag. Also edited the Dockerfile so that bootstrap doesn't need to be downloaded (it's baked into the image).

Added docs for the properties file feature.


conf/spark-env.sh
```diff
@@ -43,8 +42,6 @@ export LIBPROCESS_SSL_KEY_FILE=/mnt/mesos/sandbox/.ssl/scheduler.key
 export MESOS_MODULES="{\"libraries\": [{\"file\": \"libdcos_security.so\", \"modules\": [{\"name\": \"com_mesosphere_dcos_ClassicRPCAuthenticatee\"}]}]}"
 export MESOS_AUTHENTICATEE="com_mesosphere_dcos_ClassicRPCAuthenticatee"
 
-echo "spark-env: Environment" >&2
-env >&2
```
susanxhuynh: Why did you remove this?

ArtRand: I added that when I was debugging strict mode. I don't think a customer would necessarily want to spew the environment to the log every job.

PR #182, opened 2017-09-11
[SPARK-529] allow for application args to have arguments withou equals sign, test

ArtRand:
We were not allowing for application flags to have values. I fixed this by checking whether the following argument starts with dashes and including it in the application arguments if so. I modified `S3Job` to test this functionality and changed the test.


docs/run-job.md
```diff
@@ -8,15 +8,14 @@ enterprise: 'no'
     to a location visible to the cluster (e.g., HTTP, S3, or HDFS). [Learn more][13].
 
 1.  Run the job.
+    Include all configuration flags before the jar url and the args for your spark job after the jar url. Generally following the template `dcos spark run --submit-args="<flags> URL [args]` where `<flags>` can be things like `--conf spark.cores.max=16` and `--class my.aprk.App`, `URL` is the location of the application, and `[args]` are any arguments for the application.
+        
+        dcos spark run --submit-args=--class MySampleClass http://external.website/mysparkapp.jar"
 
-        dcos spark run --submit-args=`--class MySampleClass http://external.website/mysparkapp.jar 30`
```
susanxhuynh: Why did you remove the example that has an application arg?

tests/jobs/scala/src/main/scala/S3Job.scala
```diff
@@ -1,22 +1,42 @@
...
+      var countonly: Boolean = false
+    }
 
-    val readURL = args(0)
-    val writeURL = args(1)
-    println(s"Reading from ${readURL}.  Writing to ${writeURL}.")
+    val parser = new OptionParser[Unit]("S3 Job") {
+      opt[String]("readUrl").action((x, _) => config.readurl = x)
+      opt[String]("writeUrl").action((x, _) => config.writeurl = x)
+      opt[Unit]("countOnly").action((_, _) => config.countonly = true)
```
susanxhuynh: I don't see `countOnly` used in this job. Is that intentional?

docs/run-job.md
ArtRand: I just thought it didn't make any sense given that `http://external.website/mysparkapp.jar` isn't a real thing.

tests/jobs/scala/src/main/scala/S3Job.scala
ArtRand: It's used in the test (https://github.com/mesosphere/spark-build/blob/spark-529-scopt-args-need-equals-sign/tests/test_spark.py#L273) as a boolean flag with no args.

susanxhuynh: OK

PR #183, opened 2017-09-11
[SPARK-512] expose isR and isPython spark run args

ArtRand:
Add `--isR` and `--isPython` boolean flags to the command line to (optionally) override internal parsing of the application URL.


ArtRand:
I don't think it is exposed as a `spark-submit` option. I thought it would be useful for debugging and people who have an application without the (previously) required suffix.

PR #188, opened 2017-09-22
[SPARK-542] Add and document Driver<->Executor TLS support.

ArtRand:
## TLS in Spark

This PR is about 2 things. 
1. Update documentation and expectations surrounding TLS in Spark right now. 
2. Propose a temporary solution. 

### How it works
Spark has native support for TLS with java KeyStores. It requires the KeyStore and optionally the TrustStore (hereafter referred to as _Stores_) to be in the sandbox of the Driver _and_ the Executors. 

The solution implemented here uses environment based secrets. The user base64 encodes their Stores and puts them in the DC/OS secret store. Then you use add the appropriate configuration (either on the command line or with a properties file) and the base64 string is injected in the Driver and executor environments. The pre-start script `spark-env.sh` notices these environment variables and decodes them to files. 

### The question.
This solution is similar to the older Kerberos implementation, which is why I asked @triclambert and @vishnu2kmohan to review this PR. My question is: **is this solution inherently insecure and thus we should not even allow/advise customers to use it**? Is exposing the Stores files in the sandbox a common practice? 

The better alternative is to use file-based secrets as we do for Kerberos. Unfortunately, the MesosCoarseGrainedSchedulerBackend (the Driver) does not have secrets support which it needs to to start the executors with secrets. I recently added secret support to the Dispatcher (in Spark) and it would be a trivial change to add it to the Driver also (but would require the ASF review process). 

Thanks for the input.


conf/spark-env.sh
```diff
@@ -44,6 +44,12 @@ export MESOS_AUTHENTICATEE="com_mesosphere_dcos_ClassicRPCAuthenticatee"
 
 echo "spark-env: User: $(whoami)" >&2
 
+if base64 --help | grep -q GNU; then
+      BASE64_D="base64 -d" # GNU
+  else
+      BASE64_D="base64 -D" # BSD
+fi
+
```
skonto: How about base64 --decode, should be the same.

skonto:
@ArtRand I think sandbox access is protected https://github.com/apache/mesos/blob/master/include/mesos/authorizer/acls.proto#L193
yet depending on different envs it would be good to support both options. For example with dc/os you have better handling of env secrets using the secret store or some other implementation eg. vault.

docs/security.md
```diff
@@ -37,26 +40,35 @@ Both stores must be base64 encoded, for example:
 
 **Note:** The base64 string of the keystore will probably be much longer than the snippet above, spanning 50 lines or so.
 
-With this and the password `secret` for the keystore and the private key, your JSON options file will look like this:
+Add the stores to your secrets in the DC/OS Secret store, for example if your base64 encoded keystores and truststores are server.jks.base64 and trust.jks.base64, respectively then do the following:
```
sascala: Add the stores to your secrets in the DC/OS secret store. For example, if your base64-encoded keystores and truststores are `server.jks.base64` and `trust.jks.base64`, respectively, then use the following commands to add them to the secret store:

docs/security.md
```diff
@@ -37,26 +40,35 @@ Both stores must be base64 encoded, for example:
...
-      "security": {
-        "ssl": {
-          "enabled": true,
-          "keyStoreBase64": "/u3+7QAAAAIAAAACAAAAAgA...”,
-          "keyStorePassword": "secret",
-          "keyPassword": "secret"
-        }
-      }
-    }
+In this case you're adding two secrets `/truststore` and `/keystore` that you will need to pass to the Spark Driver and Executors. You will need to add the following configurations to your `dcos spark run ` command:
```
sascala: In this case, you are adding....

You must add the following configurations...

docs/security.md
```diff
@@ -37,26 +40,35 @@ Both stores must be base64 encoded, for example:
...
+--conf spark.mesos.task.labels=DCOS_SECRETS_DIRECTIVE:[{\"name\"\:\"/keystore\"\,\"type\"\:\"ENVIRONMENT\"\,\"environment\"\:{\"name\"\:\"KEYSTORE_BASE64\"}}\,{\"name\"\:\"/truststore\"\,\"type\"\:\"ENVIRONMENT\"\,\"environment\"\:{\"name\"\:\"TRUSTSTORE_BASE64\"}}],DCOS_SPACE:/spark, \
+--class <Spark Main class> <Spark Application JAR> [application args]"
+```
 
-Make sure to connect the DC/OS cluster only using an SSL connection (i.e. by using an `https://<dcos-url>`). Use the following command to set your DC/OS URL:
+Importantly the `spark.mesos.driver.labels` and `spark.mesos.task.labels` must be set as shown. If you upload your secret with another path (e.g. not `/keystore` and `/truststore`) then change the `name` in the value accordingly. Lastly, `spark.mesos.task.lables` must have the `DCOS_SPACE:<dcos_space>` label as well, to have access to the secret. See the [Secrets Documentation about SPACES][13] for more details about Spaces, but usually you want `/spark` as shown.
```
sascala: **Note:** The `spark.mesos.driver.labels` and `spark.mesos.task.labels` must be set as shown. If you upload your secret with another path (e.g. not `/keystore` and `/truststore`), then change the `name` in the value accordingly. Lastly, `spark.mesos.task.lables` must have also have the `DCOS_SPACE:<dcos_space>` label in order to access the secret. See the [Secrets Documentation about spaces][13] for more details about spaces. Usually, you will want to set the space label to `/spark`, as shown.

docs/security.md
```diff
@@ -37,26 +40,35 @@ Both stores must be base64 encoded, for example:
...
 
-    dcos config set core.dcos_url https://<dcos-url>
 
  [11]: https://docs.mesosphere.com/1.9/overview/architecture/components/
  [12]: http://docs.oracle.com/javase/8/docs/technotes/tools/unix/keytool.html
+ [13]: https://docs.mesosphere.com/service-docs/spark/v2.0.1-2.2.0-1/run-job/
```
susanxhuynh: As referenced above, this link is supposed to point to "Secrets Documentation about SPACES".

docs/security.md
```diff
@@ -15,19 +15,22 @@ Follow these instructions to [authenticate in strict mode](https://docs.mesosphe
...
 *   From the drivers to their executors.
 
-There are a number of configuration variables relevant to SSL setup. List them with the following command:
+There are a number of configuration variables relevant to SSL setup. The required configuration settings are:
 
-    dcos package describe spark --config
+| Variable                         | Description                                     |
+|----------------------------------|-------------------------------------------------|
+| `spark.ssl.enabled`              | Whether to enable SSL (default: `false`).       |
+| `spark.ssl.keyStoreBase64`       | Base64 encoded blob containing a Java keystore. |
```
susanxhuynh: I think this is not required anymore, if using secrets.

docs/security.md
```diff
@@ -37,26 +40,35 @@ Both stores must be base64 encoded, for example:
...
+--conf spark.ssl.enabled=true \
+--conf spark.ssl.enabledAlgorithms=TLS_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_256_CBC_SHA \
+--conf spark.ssl.keyPassword=<key password> \
+--conf spark.ssl.keyStore=server.jks \  # This MUST be set this way
+--conf spark.ssl.keyStorePassword=<keystore access password> \
+--conf spark.ssl.protocol=TLS \
+--conf spark.ssl.trustStore=trust.jks \  # this MUST be set this way
+--conf spark.ssl.trustStorePassword=<truststore password> \
+--conf spark.mesos.driver.labels=DCOS_SECRETS_DIRECTIVE:[{\"name\"\:\"/keystore\"\,\"type\"\:\"ENVIRONMENT\"\,\"environment\"\:{\"name\"\:\"KEYSTORE_BASE64\"}}\,{\"name\"\:\"/truststore\"\,\"type\"\:\"ENVIRONMENT\"\,\"environment\"\:{\"name\"\:\"TRUSTSTORE_BASE64\"}}] \
+--conf spark.mesos.task.labels=DCOS_SECRETS_DIRECTIVE:[{\"name\"\:\"/keystore\"\,\"type\"\:\"ENVIRONMENT\"\,\"environment\"\:{\"name\"\:\"KEYSTORE_BASE64\"}}\,{\"name\"\:\"/truststore\"\,\"type\"\:\"ENVIRONMENT\"\,\"environment\"\:{\"name\"\:\"TRUSTSTORE_BASE64\"}}],DCOS_SPACE:/spark, \
```
susanxhuynh: This is a pretty long command. Did you consider setting some of these configs automatically in the CLI? Or, do you want the user to be aware of each config being set?

conf/spark-env.sh
```diff
@@ -53,17 +59,28 @@ if ls ${MESOS_SANDBOX}/*.base64 1> /dev/null 2>&1; then
...
-    fi
     echo "spark-env: Copying krb config from $KRB5_CONFIG_BASE64 to /etc/" >&2
     echo "${KRB5_CONFIG_BASE64}" | ${BASE64_D} > /etc/krb5.conf
 else
     echo "spark-env: No kerberos KDC config found" >&2
 fi
 
+if [[ -n "${KEYSTORE_BASE64}" ]]; then
+    echo "spark-env: Copying keystore to SANDBOX" >&2
+    echo "${KEYSTORE_BASE64}" | ${BASE64_D} > ${MESOS_SANDBOX}/server.jks
```
mhrabovcin: I'd suggest using file based secrets that are more secure than writing contents of keystore to mesos sandbox.

susanxhuynh:
This will be replaced with file-based secrets.

susanxhuynh:
I'm going to reopen this PR and replace the env-based secrets with file-based secrets.

susanxhuynh:
I have replaced the env-based secrets with file-based secrets, pending this Apache Spark PR: https://github.com/apache/spark/pull/19437
This PR contains the related documentation and an integration test.
cc @ArtRand @skonto @mhrabovcin @sascala

tests/test_spark.py
```diff
@@ -318,6 +318,53 @@ def test_secrets():
...
+    resources_folder = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), 'resources'
+    )
+    keystore_file = 'server.jks'
+    truststore_file = 'trust.jks'
+    keystore_path = os.path.join(resources_folder, '{}.base64'.format(keystore_file))
+    truststore_path = os.path.join(resources_folder, '{}.base64'.format(truststore_file))
+    keystore_secret = '__dcos_base64__keystore'
+    truststore_secret = '__dcos_base64__truststore'
+    shakedown.run_dcos_command('security secrets create /{} --value-file {}'.format(keystore_secret, keystore_path))
```
ArtRand: how come this works now? I remember having to jump through some hoops to get the secrets API to work (https://github.com/mesosphere/spark-build/blob/master/tests/utils.py#L179). Could you homogenize the way secrets are added in the tests by either using the SecretsHandler or changing/removing it?

tests/test_spark.py
```diff
@@ -318,6 +318,53 @@ def test_secrets():
         LOGGER.warn("Error when deleting secret, {}".format(r.content))
 
 
+# Skip DC/OS < 1.10, because it doesn't have support for file-based secrets.
+@pytest.mark.skip("Enable when SPARK-22131 is merged and released in DC/OS Spark")
```
ArtRand: I assume you've tested this locally. How about we test with a specific docker image (that has the patch, built master) instead of the universe package. I don't like the idea of merging "skipped tests"

docs/security.md
```diff
@@ -15,48 +15,73 @@ Follow these instructions to [authenticate in strict mode](https://docs.mesosphe
...
+--conf spark.mesos.containerizer=mesos \  # use mesos containerizer
+--conf spark.ssl.enabled=true \
+--conf spark.ssl.enabledAlgorithms=TLS_RSA_WITH_AES_128_CBC_SHA256 \
+--conf spark.ssl.keyPassword=<key password> \
+--conf spark.ssl.keyStore=server.jks \  # This MUST be set this way
+--conf spark.ssl.keyStorePassword=<keystore access password> \
+--conf spark.ssl.protocol=TLS \
+--conf spark.ssl.trustStore=trust.jks \  # this MUST be set this way
+--conf spark.ssl.trustStorePassword=<truststore password> \
+--conf spark.mesos.driver.secret.names=__dcos_base64__keystore,__dcos_base64__truststore \
```
ArtRand: @keithchambers should probably review this. We could add some sugar to the CLI so these extra configs aren't required (like we do with Kerberos), but it becomes the same thing.

docs/security.md
```diff
@@ -15,48 +15,73 @@ Follow these instructions to [authenticate in strict mode](https://docs.mesosphe
...
+--conf spark.mesos.driver.secret.filenames=server.jks,trust.jks \
+--conf spark.mesos.executor.secret.names=__dcos_base64__keystore,__dcos_base64__truststore \
+--conf spark.mesos.executor.secret.filenames=server.jks,trust.jks \
+--conf spark.mesos.task.labels=DCOS_SPACE:/spark \
```
ArtRand: we should probably add this to the CLI automatically.

keithchambers: Would it be more intuitive for the secret:file pairs to be together?
e.g., `conf spark.mesos.executor.secrets=secret1:file1,secret2:file2`

docs/security.md
```diff
@@ -15,48 +15,67 @@ Follow these instructions to [authenticate in strict mode](https://docs.mesosphe
...
-          "enabled": true,
-          "keyStoreBase64": "/u3+7QAAAAIAAAACAAAAAgA...”,
-          "keyStorePassword": "secret",
-          "keyPassword": "secret"
-        }
-      }
-    }
+In this case, you are adding two secrets `/truststore` and `/keystore`. 
+You must add the following configurations to your `dcos spark run ` command.
+The ones in parentheses are optional.:
```
susanxhuynh: Good point, @keithchambers . Actually, that was the old verbose command. Here is the simplified one. What do you think? --

susanxhuynh:
@ArtRand I've simplified the interface a bit. There are only three required properties now, and some optional ones. PTAL.

susanxhuynh:
@ArtRand 
- Decided to go with "magic" options for passwords. I figured since any user will have to figure out our DC/OS-specific `--keystore-secret-path` option, it's not too much trouble to add the password ones also.
- Made sure any additional user-defined secrets are preserved, added placeholders for env-based secrets.
- Waiting for tests to pass one last time.

cli/dcos-spark/submit_builder.go
```diff
@@ -620,6 +689,27 @@ func buildSubmitJson(cmd *SparkCommand) (string, error) {
 		}
 	}
 
+	// TLS configuration
+	if args.keystoreSecretPath != "" {
+		// Make sure passwords are set
+		if args.keystorePassword == "" || args.privateKeyPassword == "" {
+			return "", errors.New("Need to provide keystore password and key password with keystore")
+		} else {
+			args.properties["spark.ssl.keyStorePassword"] = args.keystorePassword
```
ArtRand: nit/question: any reason the password configs are set here instead of in `setupTLSArgs?

tests/jobs/python/pi_with_secret.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+if __name__ == "__main__":
+    """
+        Usage: pi [partitions] [secret] [secret content]
+        Checks for the given env-based and file-based driver secret.
+        Then calculates the value of pi.
+    """
+
+    check_secret(sys.argv[2], sys.argv[3])
```
ArtRand: does this also check the secret in the executors or just the driver?

cli/dcos-spark/submit_builder.go
susanxhuynh: No particular reason. I'll move the setting into `setupTLSArgs` and just leave the validation check here.

tests/jobs/python/pi_with_secret.py
susanxhuynh: I didn't bother adding a test for executor secrets because the CLI code (wrt TLS) is the same for driver and executor secrets.

docs/security.md
```diff
@@ -15,48 +15,78 @@ Follow these instructions to [authenticate in strict mode](https://docs.mesosphe
...
+--keystore-password=<password to keystore> \
+--private-key-password=<password to private key in keystore> \
+(—-truststore-secret-path=<path/to/truststore, e.g. __dcos_base64__truststore> \)
+(--truststore-password=<password to truststore> \)
+(—-conf spark.ssl.enabledAlgorithms=<cipher, e.g., TLS_RSA_WITH_AES_128_CBC_SHA256> \)
+--class <Spark Main class> <Spark Application JAR> [application args]"
+```
 
-    dcos package install --options=options.json spark
+**Note:** If you have specified a space for your secrets other than the default value,
```
ArtRand: I think you can specify this with a "magic" flag? https://github.com/mesosphere/dcos-commons/pull/1872

docs/security.md
```diff
@@ -15,48 +15,78 @@ Follow these instructions to [authenticate in strict mode](https://docs.mesosphe
...
+`/spark`, then you must set `spark.mesos.task.labels=DCOS_SPACE:<dcos_space>`
+in the command above in order to access the secrets.
+See the [Secrets Documentation about spaces][13] for more details about spaces.
 
-Make sure to connect the DC/OS cluster only using an SSL connection (i.e. by using an `https://<dcos-url>`). Use the following command to set your DC/OS URL:
+**Note:** If you specify environment-based secrets with `spark.mesos.[driver|executor].secret.envkeys`,
+the keystore and truststore secrets will also show up as environment-based secrets,
+due to the way secrets are implemented. You can ignore these extra environment variables.
```
ArtRand: This should also be mentioned in Limitations (https://docs.mesosphere.com/service-docs/spark/2.1.0-2.2.0-1/limitations/)

tests/test_spark.py
```diff
@@ -286,5 +290,51 @@ def test_cli_multiple_spaces():
                           " --class  ", "org.apache.spark.examples.SparkPi"])
 
 
+# Skip DC/OS < 1.10, because it doesn't have support for file-based secrets.
+@pytest.mark.skipif('shakedown.dcos_version_less_than("1.10")')
```
ArtRand: Need to skip on open also.

docs/security.md
susanxhuynh: Which magic flag? The link ^^ doesn't seem relevant?

tests/test_spark.py
```diff
@@ -286,5 +290,51 @@ def test_cli_multiple_spaces():
...
+@pytest.mark.sanity
+def test_driver_executor_tls():
```
ArtRand: Could we make the job rely on a file being served by the driver (i.e. the channel effected by TLS)? This may be more work than necessary because we'd need to enable/check the `--files` flag.

cli/dcos-spark/submit_builder.go
```diff
@@ -257,6 +279,44 @@ func setupKerberosAuthArgs(args *sparkArgs) error {
...
+	}
+	joinedPaths := strings.Join(paths, ",")
+	joinedFilenames := strings.Join(filenames, ",")
+	joinedEnvkeys := strings.Join(envkeys, ",")
+
+	taskTypes :=[]string{"driver", "executor"}
+	for _, taskType := range taskTypes {
+		appendToProperty(fmt.Sprintf("spark.mesos.%s.secret.names", taskType), joinedPaths, args)
+		appendToProperty(fmt.Sprintf("spark.mesos.%s.secret.filenames", taskType), joinedFilenames, args)
+		appendToPropertyIfSet(fmt.Sprintf("spark.mesos.%s.secret.envkeys", taskType), joinedEnvkeys, args)
```
ArtRand: :+1:

tests/jobs/python/pi_with_secret.py
```diff
@@ -0,0 +1,48 @@
+import os
+import sys
+from random import random
+from operator import add
+
+from pyspark.sql import SparkSession
+
+def check_secret(secret_name, secret_content):
+    '''
+    Make sure the extra secret envvar and secret file show up in driver.
```
ArtRand: check driver and executor?

docs/security.md
susanxhuynh: Adding ...

tests/test_spark.py
susanxhuynh: Added

PR #189, opened 2017-09-25
[SPARK-370][SPARK-499] Change image to run as nobody

ArtRand:
## Problem
The Spark image that is used for running the Dispatcher, Drivers, and Executors runs as `root` meaning that Marathon requires permissions to launch containers as `root` when on a strict-mode cluster. 

## Solution
Change default to `nobody`, allow access to directories in the image necessary for running Spark and it's components (e.g. nginx).

## Tested
CI unit tests for basic functionality, manually on a strict-mode cluster (see [JIRA](https://jira.mesosphere.com/browse/SPARK-370) for details) and manually against a Kerberized HDFS cluster

### Manual tests
- [x] Kerberos
- [x] History server


docker/Dockerfile
```diff
@@ -72,15 +72,18 @@ RUN ln -s /bin/grep /usr/bin/grep
 
 RUN ln -s /var/lib/runit/service/spark /etc/service/spark
 RUN ln -s /var/lib/runit/service/nginx /etc/service/nginx
+RUN mkdir -p /etc/hadoop
```
susanxhuynh: This is repeated in L. 60 above, one of them can be removed.

docker/Dockerfile
```diff
@@ -76,11 +76,13 @@ RUN ln -s /var/lib/runit/service/nginx /etc/service/nginx
 ADD dist /opt/spark/dist
 
 # Commenting these for now, because we're running Spark as root in strict mode
```
benclarkwood: What does this comment mean?

benclarkwood: It seems like it should go away?

skonto:
I am not so sure about nobody in general its a relic: https://lwn.net/Articles/695526/

PR #193, opened 2017-10-16
Spark dcos CLI breaks with multiple spaces

skonto:
Spaces between arguments break the spark cli:
https://github.com/mesosphere/spark-build/blob/master/cli/dcos-spark/submit_builder.go#L302

This will fail when the job is run:
```
dcos spark run --verbose --submit-args="--conf spark.streaming.fileStream.minRememberDuration=1200s__--driver-cores 1 --conf spark.cores.max=4 --driver-memory 1024M --class com.lightbend.fdp.spark.test.FileStreamWithCheckpointing https://s3-eu-west-1.amazonaws.com/fdp-stavros-test/checkpointing-test-assembly-1.0-SNAPSHOT.jar --checkpointDirectory /checkpoint --outputDirectory /output --inputDirectory /watched"dcos spark run --verbose --submit-args="--conf spark.streaming.fileStream.minRememberDuration=1200s --driver-cores 1 --conf spark.cores.max=4 --driver-memory 1024M --class com.lightbend.fdp.spark.test.FileStreamWithCheckpointing https://s3-eu-west-1.amazonaws.com/fdp-stavros-test/checkpointing-test-assembly-1.0-SNAPSHOT.jar --checkpointDirectory /checkpoint --outputDirectory /output --inputDirectory /watched"
```

Error is not informative on mesos:
I1016 10:31:50.852361  7563 fetcher.cpp:285] Fetching directly into the sandbox directory
I1016 10:31:50.852404  7563 fetcher.cpp:222] Fetching URI ''
E1016 10:31:50.852429  7563 fetcher.cpp:579] EXIT with status 1: Failed to fetch '': A relative path was passed for the resource but the Mesos framework home was not specified. Please either provide this config option or avoid using a relative path

There are two spaces between those args: --conf spark.streaming.fileStream.minRememberDuration=1200s --driver-cores 1 --conf. I am using underscore to show that.

Another case is when we have:
`--conf    spark.app.name=kerberosStreaming`
Two spaces again.
This will make dcos cli fail immediately:
2017/10/16 13:33:40 Error when parsing --submit-args: expected KEY=VALUE got ''

### DIFF
--- a/cli/bin/build-go.sh
+++ b/cli/bin/build-go.sh
@@ -34,7 +34,7 @@ export GOPATH=$(pwd)/vendor:$GOPATH
 
 # this may be omitted in 1.6+, left here for compatibility with 1.5:
 export GO15VENDOREXPERIMENT=1
-
+go test
 # available GOOS/GOARCH permutations are listed at:
 # https://golang.org/doc/install/source#environment
 
--- a/cli/dcos-spark/cli_test.go
+++ b/cli/dcos-spark/cli_test.go
@@ -1,0 +1,39 @@
+package main
+
+import "testing"
+
+// test spaces
+func TestCleanUpSubmitArgs(t *testing.T) {
+  _, args := sparkSubmitArgSetup()
+  inputArgs := "--conf    spark.app.name=kerberosStreaming   --conf spark.cores.max=8"
+  submitArgs, _ := cleanUpSubmitArgs(inputArgs, args.boolVals)
+    if "--conf=spark.app.name=kerberosStreaming" != submitArgs[0] {
+       t.Errorf("Failed to reduce spaces while cleaning submit args.")
+    }
+
+    if "--conf=spark.cores.max=8" != submitArgs[1] {
+       t.Errorf("Failed to reduce spaces while cleaning submit args.")
+    }
+}
+
+// test scopts pattern for app args when have full submit args
+func TestScoptAppArgs(t *testing.T) {
+  _, args := sparkSubmitArgSetup()
+  inputArgs := `--driver-cores 1 --conf spark.cores.max=1 --driver-memory 512M
+  --class org.apache.spark.examples.SparkPi http://spark-example.jar --input1 value1 --input2 value2`
+  submitArgs, appFlags := cleanUpSubmitArgs(inputArgs, args.boolVals)
+
+  if "--input1" != appFlags[0] {
+     t.Errorf("Failed to parse app args.")
+  }
+  if "value1" != appFlags[1] {
+     t.Errorf("Failed to parse app args.")
+  }
+
+  if "--driver-memory=512M" != submitArgs[2] {
+     t.Errorf("Failed to parse submit args..")
+  }
+  if "http://spark-example.jar" != submitArgs[4] {
+     t.Errorf("Failed to parse submit args..")
+  }
+}
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -20,6 +20,7 @@ import (
 
 var keyWhitespaceValPattern = regexp.MustCompile("(.+)\\s+(.+)")
 var backslashNewlinePattern = regexp.MustCompile("\\s*\\\\s*\\n\\s+")
+var collapseSpacesPattern = regexp.MustCompile(`[\s\p{Zs}]{2,}`)
 
 type sparkVal struct {
 	flagName string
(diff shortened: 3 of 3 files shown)


skonto:
@susanxhuynh @ArtRand pls review.

cli/dcos-spark/submit_builder.go
```diff
@@ -298,8 +299,11 @@ func parseApplicationFile(args *sparkArgs) error {
...
 func cleanUpSubmitArgs(argsStr string, boolVals []*sparkVal) ([]string, []string) {
+
+  // collapse two or more spaces to one.
+	argsCompacted := collapseSpacesPattern.ReplaceAllString(argsStr, " ")
 	// clean up any instances of shell-style escaped newlines: "arg1\\narg2" => "arg1 arg2"
-	argsCleaned := strings.TrimSpace(backslashNewlinePattern.ReplaceAllLiteralString(argsStr, " "))
+	argsCleaned := strings.TrimSpace(backslashNewlinePattern.ReplaceAllLiteralString(argsCompacted, " "))
 	// HACK: spark-submit uses '--arg val' by convention, while kingpin only supports '--arg=val'.
 	//       translate the former into the latter for kingpin to parse.
 	args := strings.Split(argsCleaned, " ")
```
elezar: Another option would be to split `argsCleaned` using a `Regexp.Split` instead of `strings.Split`:
```go
package main

import (
	"fmt"
	"regexp"
)

func main() {
	argString := "one two  three   four"

	splitter := regexp.MustCompile(`\s+`)
	for _, s := range splitter.Split(argString, -1) {
		fmt.Println(s)
	}
}
```

It would also be nice to add `submit_builder_test.go` with some unit tests for the various cases here.

skonto: Yes there are multiple option to fix this I guess. I agree we need tests since this is not the only issue found so far (probably will need to address it in a separate PR). @susanxhuynh are you runnign any tests for this internally?

elezar: As far as I know, there are no additional tests for this kind of behaviour, and adding a simple unit tests here would add great value. If you don't get to it, I will see if I can add something this weekend.

skonto: @elezar ok I will check what I can add here.

skonto:
@susanxhuynh @elezar I added a basic unit test. @susanxhuynh Let me know if I need to trigger testing in more places. Is the setup ok for jenkins?

cli/bin/test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set +x
+
+BIN_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+cd "${BIN_DIR}/.."
+
+if [ -z "$GOPATH" -o -z "$(which go)" ]; then
```
skonto: This part is also found in build-go.sh and it is repeated here. I don't like bash include files but maybe we should add it in one. Let's verify that this works for the jenkins stuff.

skonto:
@susanxhuynh could you assist with failures and how to integrate with the jenkins setup.

susanxhuynh:
Yeah, I'll check why the jenkins build failed, and figure out how to run the test in CI.

susanxhuynh:
retest this please

cli/bin/test.sh
ArtRand: I would almost prefer that we don't add it to `build-go.sh` so that you can build the CLI without invoking the tests (or commenting out lines then forgetting to uncomment them). FWIW I usually build the CLI then just copy it to my `.dcos/` dir for testing the CLI.

cli/dcos-spark/cli_test.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+import "testing"
+
+// verify kingpin gets --arg=val
```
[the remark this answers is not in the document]
ArtRand: would it also be hard to add a test for `scopt` args?

cli/dcos-spark/submit_builder.go
ArtRand: The CLI is "tested" by the integration tests, they will fail fast if it's not working. Simple unit tests are fine, but IMO opinion we could just add to the CI tests given that the unit tests would be very simple (akin to getter/setter basically).

cli/dcos-spark/cli_test.go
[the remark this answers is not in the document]
skonto: Ok will add a test sure.

skonto:
@susanxhuynh I updated the PR.

cli/bin/build-go.sh
```diff
@@ -57,3 +57,5 @@ case "$OSTYPE" in
     linux*) strip "${EXE_NAME}${SUFFIX}"
 esac
 print_file "${EXE_NAME}${SUFFIX}"
+
+go test
```
susanxhuynh: @skonto Could you add this earlier, like around L. 37? So the logic would be: if the unit tests do not pass, we do not bother building the (Windows, OSX, Linux) executables.

skonto: Ok I thought you build first and then test, sure.

skonto: Done.

skonto: @susanxhuynh I dont need rebasing I think I touched different files.

susanxhuynh: @skonto My recent fix is needed to get one of the CI tests to pass (something changed in hdfs and I had to add a workaround). So, I think you have to include that fix in order for the Jenkins tests to pass for this PR.

susanxhuynh:
@skonto Only thing missing is rebasing to pull in my CI fix. Otherwise, CI test ("test:shakedown") will not pass in this PR.

skonto:
@susanxhuynh done

PR #195, opened 2017-10-23
Temporarily pinned the hdfs version to get 'test_s3' CI test passing

susanxhuynh:
There's a newly introduced bug in the latest version of hdfs in the Universe (2.0.2-xyz), in one of the s3n config values. This prevents Spark from reading from S3 when it is configured to run with HDFS. To get the CI tests passing again, I have temporarily pinned the version of HDFS to the previous version (2.0.1-xyz).

Can unpin when the hdfs config bug is fixed.


susanxhuynh:
cc @ArtRand

PR #202, opened 2017-11-06
Updated JRE version to 8u152

susanxhuynh:


ArtRand:
I'd like to test locally, but I'm sure this is fine. Did you test with Kerberos?

susanxhuynh:
No, I haven't tested with Kerberos - I'll do so when I get a Kerberized HDFS.

susanxhuynh:
Tested reading and writing to a Kerberized HDFS (built from dcos-commons).

PR #207, opened 2017-11-20
[SPARK-570] CLI does not allow long strings of options, e.g. spark.driver.extraJavaOpts

ArtRand:
It's common that someone wants to set multiple java options `-Djava.setting=parameter`. Right now we only let one setting at a time, and extra arguments starting with `-D` will result in an unknown argument error. In the CLI _and_ `spark-submit` if the same configuration is set multiple times only the last setting is taken - we retain this functionality. We use single quotes `'` to designate the start of a string of arguments. 

Example:
```bash
dcos spark run --submit-args="\
...
--conf spark.driver.extraJavaOptions='-Djava.something=somethingElse -Djava.middleConfig=middleSetting -Djava.parameter=setting' \
--conf spark.executor.extraJavaOptions='-Djava.something=somethingElse -Djava.parameter=setting' \
...
```

or 

```bash
dcos spark run --submit-args="\
--driver-java-options='-Djava.something=somethingelse -Djava.parameter=setting' \
...
```

### DIFF
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -313,6 +313,7 @@ func cleanUpSubmitArgs(argsStr string, boolVals []*sparkVal) ([]string, []string
 	argsEquals := make([]string, 0)
 	appFlags := make([]string, 0)
 	i := 0
+	inQuotes := false
 ARGLOOP:
 	for i < len(args) {
 		arg := args[i]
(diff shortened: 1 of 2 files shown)


cli/dcos-spark/submit_builder.go
```diff
@@ -354,12 +363,36 @@ ARGLOOP:
...
+			if strings.HasPrefix(next, "'") {  // e.g. --driver-java-options '-Djava.config=setting... <-- next
+				inQuotes = true
+			}
+			next = strings.Replace(next, "'", "", -1)  // remove internal quotes
+			argsEquals = append(argsEquals, arg + "=" + next)
 			i += 2
+		} else if strings.HasSuffix(arg, "'") {  // attach the final arg to the string of args without the quote
+			inQuotes = false  // has suffix means we're out of the quotes
+			arg = strings.TrimSuffix(arg, "'")
+			argsEquals[len(argsEquals) - 1] = argsEquals[len(argsEquals) - 1] + " " + arg
```
susanxhuynh: I think you could use "+=" here.

elezar:
Ticket has been closed as "Won't do" closing PR as stale.

PR #214, opened 2017-11-27
[SPARK-575] Modified Soak test "delete-terasort-files" job for Kerberized HDFS

susanxhuynh:
In order to work against a Kerberized HDFS, a new "hdfs-kerberos-delete-terasort-files" job does a `dcos task exec` into a `hdfsclient` Marathon app. It runs `kinit` before running HDFS dfs commands.

- Also, made the Spark service name configurable from the environment. There may be multiple instances of Spark running in the soak cluster.


tests/test_soak.py
```diff
@@ -10,7 +10,7 @@
 
 
 LOGGER = logging.getLogger(__name__)
-SOAK_SPARK_APP_NAME='/spark'
+SOAK_SPARK_SERVICE_NAME = os.getenv('SOAK_SPARK_SERVICE_NAME', '/spark')
```
ArtRand: this is un utils.py already isn't it?

susanxhuynh: @ArtRand No, only the package name is in utils.py. The service name varies by test, so it's not normally read from the environment. I think only in the soak case do we want to read the service name from the environment (because in the soak cluster we're not constantly installing and uninstalling Spark).

PR #218, opened 2017-11-29
[SPARK-559] Changes for strict mode CI tests

susanxhuynh:
Changes for the tests to run in strict mode
* Makefile change to spin up a strict cluster
* Setup service account, grant permissions
** Moved from old `setup_permissions.sh` script to sdk_security module
* Set a default "/spark" app name; granted permissions to drivers with that app name
* Marked test_jar() and test_marathon_group(), which runs test_jar(), as “expected fail”; failures tracked by SPARK-593


susanxhuynh:
cc @ArtRand @skonto

Makefile
```diff
@@ -135,6 +135,7 @@ test-env:
 	source test-env/bin/activate
 	pip3 install -r tests/requirements.txt
 
+CF_TEMPLATE_URL ?= https://s3.amazonaws.com/downloads.mesosphere.io/dcos-enterprise/testing/master/cloudformation/ee.single-master.cloudformation.json
```
ArtRand: this is the strict CF template?

tests/test_kafka.py
```diff
@@ -140,7 +151,7 @@ def test_pipeline(kerberos_flag, stop_count, jar_uri, keytab_secret, jaas_uri=No
         "--conf", "spark.mesos.driver.secret.filenames=kafka-client.keytab",
         "--conf", "spark.mesos.executor.secret.names={}".format(keytab_secret),
         "--conf", "spark.mesos.executor.secret.filenames=kafka-client.keytab",
-        "--conf", "spark.mesos.task.labels=DCOS_SPACE:/spark",
+        "--conf", "spark.mesos.task.labels=DCOS_SPACE:{}".format(utils.SPARK_APP_NAME),
```
ArtRand: This is done [in the CLI](https://github.com/mesosphere/spark-build/blob/master/cli/dcos-spark/submit_builder.go#L651) and iff you want to specify a different DCOS_SPACE, it should be done as a [DC/OS arg](https://github.com/mesosphere/spark-build/blob/master/cli/dcos-spark/main.go#L25) - I can make the PR to change this, however.

tests/test_spark.py
```diff
@@ -187,15 +187,15 @@ def test_s3():
     utils.run_tests(app_url=utils._scala_test_jar_url(),
                     app_args=app_args,
                     expected_output="Read 3 lines",
-                    app_name="/spark",
                     args=args)
 
 
+@pytest.mark.xfail(utils.is_strict(), reason="Currently fails in strict mode")
```
ArtRand: We should change this test to something that can be exercised in `strict`, probably still good to test this as folding (marathon groups) is becoming the norm.

tests/utils.py
```diff
@@ -2,18 +2,26 @@
...
 
-import json
 import logging
 import os
 import re
-import requests
 import shakedown
 import subprocess
 import urllib
+import urllib.parse
```
ArtRand: is there a reason this can't be done with the above import?

tests/utils.py
```diff
@@ -175,11 +179,10 @@ def upload_file(file_path):
     return s3.http_url(basename)
 
 
-def submit_job(app_url, app_args, app_name="/spark", args=[]):
+def submit_job(app_url, app_args, app_name=SPARK_APP_NAME, args=[]):
     if is_strict():
-        args += ["--conf", 'spark.mesos.driverEnv.MESOS_MODULES=file:///opt/mesosphere/etc/mesos-scheduler-modules/dcos_authenticatee_module.json']
-        args += ["--conf", 'spark.mesos.driverEnv.MESOS_AUTHENTICATEE=com_mesosphere_dcos_ClassicRPCAuthenticatee']
-        args += ["--conf", 'spark.mesos.principal=service-acct']
+        args += ["--conf", 'spark.mesos.containerizer=mesos']
```
ArtRand: is MESOS containerizer required for `strict`? (wasn't in the past)

tests/utils.py
```diff
@@ -241,3 +244,55 @@ def teardown_spark():
 
 def _scala_test_jar_url():
     return s3.http_url(os.path.basename(os.environ["SCALA_TEST_JAR_PATH"]))
+
+
+def spark_security_session():
+    '''
+    Spark strict mode setup is slightly different from dcos-commons, so can't use sdk_security::security_session.
```
ArtRand: explain why?

tests/utils.py
```diff
@@ -241,3 +244,55 @@ def teardown_spark():
...
+    '''
+    user = 'nobody'
+    role = '*'
+    service_account = SPARK_SERVICE_ACCOUNT
+    secret = SPARK_SERVICE_ACCOUNT_SECRET
+
+    def grant_driver_permission(service_account_name, app_id):
+        dcosurl, headers = sdk_security.get_dcos_credentials()
+        app_id_encoded = urllib.parse.quote(app_id, safe='')
+        app_id_encoded = urllib.parse.quote(app_id_encoded, safe='')  # double-encoded (why?)
```
ArtRand: could you nest these so there isn't two assignments?

tests/utils.py
```diff
@@ -241,3 +244,55 @@ def teardown_spark():
 
 def _scala_test_jar_url():
     return s3.http_url(os.path.basename(os.environ["SCALA_TEST_JAR_PATH"]))
+
+
+def spark_security_session():
```
ArtRand: 👍

Makefile
susanxhuynh: This is permissive by default. The CI job in TeamCity (spark-build-PR-strict) has this set to a strict template.

tests/test_kafka.py
susanxhuynh: OK, I'll let you change it in a separate PR.

tests/test_spark.py
susanxhuynh: Yeah, guess it doesn't matter much which test it is. I'll change it to SparkPi.

tests/utils.py
susanxhuynh: If I only import `urllib`, the following line will fail:
`app_id_encoded = urllib.parse.quote(app_id, safe='')`

tests/utils.py
susanxhuynh: Let me double-check ... I don't remember why I added it.

tests/utils.py
susanxhuynh: I'll add more details about what's different: (1) the role is "*", (2) the driver permissions to execute tasks.

tests/utils.py
susanxhuynh: Sure

susanxhuynh:
@ArtRand I remember why I added mesos containerizer: it was either that or needing to set SPARK_USER in docker containerizer. I've changed it to the latter, since that's our current default (docker).

susanxhuynh:
@ArtRand Addressed your comments. PTAL.

ArtRand:
LGTM Thanks @susanxhuynh

PR #219, opened 2017-11-29
[INFINITY-2624] kerberos kafka soak

ArtRand:


tests/test_soak.py
```diff
@@ -35,6 +41,18 @@ def test_terasort():
         _run_teravalidate()
 
 
+@pytest.mark.soak
+@pytest.mark.rand
```
susanxhuynh: Was this for debugging?

tests/test_kafka.py
```diff
@@ -153,30 +149,45 @@ def kafka_broker_dns():
...
+    return utils.streaming_job_launched(PRODUCER_SERVICE_NAME)
+
+
+def _producer_started():
+    return utils.streaming_job_running(PRODUCER_SERVICE_NAME)
+
+
+def _kafka_broker_dns():
+    cmd = "{package_name} --name={service_name} endpoints broker".format(
+        package_name=KAFKA_PACKAGE_NAME, service_name=KAFKA_SERVICE_NAME)
```
susanxhuynh: The service name may be different in the soak cluster - "kafka-kerberos". Is it configurable in the test?

tests/test_soak.py
ArtRand: whoops.

tests/test_kafka.py
ArtRand: good catch, will change

PR #225, opened 2017-12-04
[SPARK-468] kerberos docs round2

ArtRand:


ArtRand:
reminder: add note that you must always provide a keytab and principal when using kerberos-hdfs

susanxhuynh:
@ArtRand Is this ready for review?

ArtRand:
@susanxhuynh yes. But there are still some `example.com` references where I need to link to SDK docs that haven't been written. In general, if there are any major mistakes or we don't like the new organization we can start working on this now.

docs/hdfs.md
```diff
@@ -16,12 +18,12 @@ To configure `hdfs.config-url` to be a URL that serves your `hdfs-site.xml` and
   }
 }
 ```
+This can also be done through the UI. If you are using the default installation of HDFS from Mesosphere this is probably `http://api.hdfs.marathon.l4lb.thisdcos.directory/v1/endpoints`.
 
-For more information, see [Inheriting Hadoop Cluster Configuration][8].
-
-For DC/OS HDFS, these configuration files are served at `http://<hdfs.framework-name>.marathon.mesos:<port>/v1/endpoints`, where `<hdfs.framework-name>` is a configuration variable set in the HDFS package, and `<port>` is the port of its marathon app.
+# Adding HDFS files per-job
+To add the configuration files manually for a job, use `--conf spark.mesos.uris=<loation_of_hdfs-site.xml>,<location_of_core-site.xml>`. This will download the files to the sandbox of the Driver Spark application, and DC/OS Spark will automatically load these files into the correct location. **Note** It is important these files are called `hdfs-site.xml` and `core-site.xml`.
```
susanxhuynh: (sp) "lo**c**ation_of_hdfs-site.xml"

docs/kerberos.md
```diff
@@ -0,0 +1,124 @@
...
+---
+
+
+# HDFS Kerberos
+
+Kerberos is an authentication system to allow Spark to retrieve and write data securely to a Kerberos-enabled HDFS cluster. As of Mesosphere Spark `2.2.0-2`, long-running jobs will renew their delegation tokens (authentication credentials). This section assumes you have previously set up a Kerberos-enabled HDFS cluster. For instruction on how to setup a secure HDFS cluster see the [HDFS documentation][http://example.com]
+
+## Spark Installation
+
+Spark (and all Kerberos-enabed) components need a valid `krb5.conf` file. You can setup the Spark service to use a single `krb5.conf` file for all of the it's drivers.
```
susanxhuynh: "all of its drivers"

docs/kerberos.md
```diff
@@ -0,0 +1,124 @@
...
+
+         {
+            "history-server": {
+                "kerberos": {
+                  "principal": "spark@REALM",
+                  "keytab": "<base64 encoding>"
+                }
+            }
+         }
+
```
susanxhuynh: Are we missing a step of adding a keytab or tgt to the secret store?

docs/kerberos.md
```diff
@@ -0,0 +1,124 @@
...
+        cat krb5.conf | base64 -w 0
+        
+    Then assign the environment variable, `KRB5_CONFIG_BASE64`, this value for the Driver and the Executors:
+        --conf spark.mesos.driverEnv.KRB5_CONFIG_BASE64=<base64_encoded_string>
+        --conf spark.executorEnv.KRB5_CONFIG_BASE64=<base64_encoded_string>
+        
+*   The `keytab` containing the credentials for accessing the Kafka cluster.
+        
+        --conf spark.mesos.driver.secret.names=<base64_encoded_keytab>    # e.g. __dcos_base64__kafka_keytab
+        --conf spark.mesos.driver.secret.filenames=kafka-client.keytab    # e.g. __dcos_base64__kafka_keytab
```
susanxhuynh: the "e.g." example looks wrong here

docs/kerberos.md
```diff
@@ -0,0 +1,124 @@
...
+        --conf spark.mesos.executor.secret.names=<base64_encoded_keytab>  # e.g. __dcos_base64__kafka_keytab
+        --conf spark.mesos.executor.secret.filenames=kafka-client.keytab  # e.g. __dcos_base64__kafka_keytab
```
susanxhuynh: Again, wrong "e.g." example

docs/kerberos.md
ArtRand: I didn't change any of the instructions w.r.t. the history server because I know that's still a WIP. But I'll add a note about uploading the keytab.

docs/kerberos.md
```diff
@@ -0,0 +1,124 @@
...
+To authenticate to a Kerberos KDC, Spark on Mesos supports keytab files as well as ticket-granting tickets (TGTs). Keytabs are valid infinitely, while tickets can expire. Keytabs are recommended, especially for long-running streaming jobs.
+
+### Keytab Authentication
+
+Submit the job with the keytab:
+
+    dcos spark run --submit-args="\
+    --kerberos-principal user@REALM \
+    --keytab-secret-path /__dcos_base64__hdfs-keytab \
+    --conf ... --class MySparkJob <url> <args>"
```
susanxhuynh: @ArtRand I don't know if we want to document it, but to get this working in Soak with CentOS, I had to add this:
`--conf spark.mesos.driverEnv.SPARK_USER=root`, to get the executors to run as `root`.

ArtRand: All of the docs relating to deployment in strict and permissive modes is here (https://github.com/mesosphere/dcos-docs-enterprise/blob/master/service-docs/spark/spark-auth.md). I would be in favor of moving all of this to it's own section. WDYT?

susanxhuynh: What are you proposing to move to its own section? The installing as `root` in CentOS? -- Since it only comes up for Kerberos, I was thinking it belonged in the `kerberos.md`.

susanxhuynh:
Aside from that question about CentOS, everything else LGTM.

docs/kerberos.md
ArtRand: I'll add a note. But I think that installing as `root` on CentOS should be in the limitations or install notes. I'll add it there. Do you have a description of the problem?

susanxhuynh: Yeah, I agree. We can copy the blurb from the Cassandra docs: https://docs.mesosphere.com/service-docs/cassandra/2.0.3-3.0.14/limitations/#service-user

ArtRand: Do you know/remember the source of the problem that required you to set `--conf spark.mesos.driverEnv.SPARK_USER=root`? I'll add the note, but I'd like to have some context.

PR #239, opened 2017-12-18
Changed history server package name to "beta-spark-history"

susanxhuynh:


susanxhuynh:
One test failure: test_rpc_auth, because it has Spark package name hard-coded. Merging anyway.

PR #249, opened 2018-01-11
[SPARK-613] Don't override explicit configuration with defaults

ArtRand:


ArtRand:
We don't have unit tests for the `buildSubmitJson` mega-function. So I checked it by eye, and the logic is pretty simple. Refactoring the `buildSubmitJson` function to be more easily tested may be justified, but I think is out of the scope of this PR given that this is a bug. Don't you think?

susanxhuynh:
OK, don't worry about it then - I didn't realize it would be so difficult to add a unit test.

PR #251, opened 2018-01-16
[SPARK-573] removed base64 decoding logic

ArtRand:
Remove ability to automatically decode files fetched with `.base64` prefix and add cli tests.


cli/dcos-spark/cli_test.go
```diff
@@ -1,39 +1,179 @@
...
+	}
+
+	//v := reflect.ValueOf(m["sparkProperties"])
+	stringProps := map[string]string{
+		"spark.driver.cores": driverCores,
+		"spark.cores.max": maxCores,
+		"spark.mesos.executor.docker.forcePullImage": "true",
+		"spark.mesos.executor.docker.image": image,
+		"spark.mesos.task.labels": fmt.Sprintf("DCOS_SPACE:%s", space),
+		"spark.ssl.noCertVerification": "true",
```
skonto: What if I set my cluster with core.ssl_verify=true? This should be false right?

PR #255, opened 2018-01-19
Make properties configurable

skonto:
@susanxhuynh I know the purpose was to hardcode some values and release. I also need this for the CI stuff I am doing. I think it does not change anything for your side :)


ArtRand:
LGTM.

PR #257, opened 2018-01-20
[SPARK-630] Align Spark config with other data services

ArtRand:
# Add config options to Spark service to align with other SDK-based data services

## Main changes
1. **BREAKING CHANGE** Setting the `krb5.conf` via the base64 encoded blob requires the user to specify `service.security.kerberos.enabled`. 
2. User can now specify `service.security.kerberos.kdc.[port|hostname]` and `service.security.kerberos.realm` directly in the config. 
3. Old behavior with the base64 encoded blob remains the same, and will overwrite the above new configs. 

## Other 
1. Move code relating to security/kerberos to it's own go file (`submit-builder.go` was getting a little ridiculous)
2. Update hdfs tests to use the new configs (implicitly tests the same thing). Also the kafka tests still use the blob.


cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+const sparkKdcHostnameKey = "SPARK_SECURITY_KERBEROS_KDC_HOSTNAME"
+const sparkKdcPortKey = "SPARK_SECURITY_KERBEROS_KDC_PORT"
+const sparkKerberosRealmKey = "SPARK_SECURITY_KERBEROS_REALM"
+const sparkKrb5ConfBase64BlobKey = "SPARK_MESOS_KRB5_CONF_BASE64"
+
+
+func prepareBase64Secret(secretPath string) string {
+	ss := strings.Split(secretPath, "/")
+	// The secret file without any slashes
+	fn := ss[len(ss) - 1]
```
elezar: If `secretPath` is empty, this will error? Should we handle this gracefully?

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+const SECRET_ENVKEY_TEMPLATE = "spark.mesos.%s.secret.envkeys"
+const sparkKdcHostnameKey = "SPARK_SECURITY_KERBEROS_KDC_HOSTNAME"
+const sparkKdcPortKey = "SPARK_SECURITY_KERBEROS_KDC_PORT"
+const sparkKerberosRealmKey = "SPARK_SECURITY_KERBEROS_REALM"
+const sparkKrb5ConfBase64BlobKey = "SPARK_MESOS_KRB5_CONF_BASE64"
+
+
+func prepareBase64Secret(secretPath string) string {
+	ss := strings.Split(secretPath, "/")
+	// The secret file without any slashes
```
elezar: This comment is unclear. Rather rename the `fn` variable to indicate what we are extracting.

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+	"errors"
+	"crypto/rand"
+	"encoding/base64"
+	"strings"
+	"log"
+)
+
+
+const KEYLENGTH = 128
+const SECRET_REFERENCE_TEMPLATE = "spark.mesos.%s.secret.names"
```
elezar: From what I can tell, there are only two options for `%s`. Instead of using `Sprintf` every time, we could use a map instead.

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+	if strings.HasPrefix(fn, "__dcos_base64__") {
+		return strings.TrimPrefix(fn, "__dcos_base64__")
+	}
+	return fn
+}
+
+func addArgsForFileBasedSecret(args *sparkArgs, secretPath, property string) {
+	taskTypes := []string{"driver", "executor"}
+	for _, taskType := range taskTypes {
+		secretRefProp := fmt.Sprintf(SECRET_REFERENCE_TEMPLATE, taskType)
```
elezar: See the comment on using a map above.

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	joinedPaths := strings.Join(paths, ",")
+	joinedFilenames := strings.Join(filenames, ",")
+	joinedEnvkeys := strings.Join(envkeys, ",")
+
+	taskTypes := []string{"driver", "executor"}
+	for _, taskType := range taskTypes {
+		appendToProperty(fmt.Sprintf(SECRET_REFERENCE_TEMPLATE, taskType), joinedPaths, args)
+		appendToProperty(fmt.Sprintf(SECRET_FILENAME_TEMPLATE, taskType), joinedFilenames, args)
+		appendToPropertyIfSet(fmt.Sprintf(SECRET_ENVKEY_TEMPLATE, taskType), joinedEnvkeys, args)
```
elezar: Are the envKeys only used for TLS?

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+	// Secret paths, filenames, and place holder envvars
+	paths := []string{args.keystoreSecretPath}
+	filenames := []string{keyStoreFileName}
+	envkeys := []string{"DCOS_SPARK_KEYSTORE"}
+	if args.truststoreSecretPath != "" {
+		paths = append(paths, args.truststoreSecretPath)
+		filenames = append(filenames, trustStoreFileName)
+		envkeys = append(envkeys, "DCOS_SPARK_TRUSTSTORE")
+	}
+	joinedPaths := strings.Join(paths, ",")
```
elezar: Instead of maintaining joined comma-separated strings, would it make sense to maintain lists for the variables and then join them once at the end?

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+
+func setupTLSArgs(args *sparkArgs) {
+	args.properties["spark.mesos.containerizer"] = "mesos"
+	args.properties["spark.ssl.enabled"] = "true"
+
+	// Keystore and truststore
+	const keyStoreFileName = "server.jks"
+	const trustStoreFileName = "trust.jks"
+	args.properties["spark.ssl.keyStore"] = keyStoreFileName
+	if args.truststoreSecretPath != "" {
```
elezar: Could this be moved into the conditional block below?

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+		appendToProperty(fmt.Sprintf(SECRET_REFERENCE_TEMPLATE, taskType), joinedPaths, args)
+		appendToProperty(fmt.Sprintf(SECRET_FILENAME_TEMPLATE, taskType), joinedFilenames, args)
+		appendToPropertyIfSet(fmt.Sprintf(SECRET_ENVKEY_TEMPLATE, taskType), joinedEnvkeys, args)
+	}
+
+	// Passwords
+	args.properties["spark.ssl.keyStorePassword"] = args.keystorePassword
+	args.properties["spark.ssl.keyPassword"] = args.privateKeyPassword
+
+	if args.truststoreSecretPath != "" {
```
elezar: It is not 100% clear what `args.truststoreSecretPath` affects. It is checked in multiple places and various options are set.

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+		args.properties["spark.ssl.trustStorePassword"] = args.truststorePassword
+	}
+
+	// Protocol
+	if _, ok := args.properties["spark.ssl.protocol"]; !ok {
```
elezar: If we don't allow other settings for `spark.ssl.protocol`, then we should simply set it here instead of only setting it when empty. If other settings are allows, one should probably log overwrites or invalid settings here.

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+func GetRandomStringSecret() (string, error) {
+	b, err := generateRandomBytes(KEYLENGTH)
+	return base64.URLEncoding.EncodeToString(b), err
+}
+
+// KERBEROS
+
+func ValidateInputs(args *sparkArgs) error {
+	if args.keytabSecretPath == "" && args.tgtSecretPath == "" && args.tgtSecretValue == "" {
+		return errors.New("need to provide Keytab secret, TGT secret, or TGT value " +
```
elezar: Should this not check the principal as well?

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	return nil
+}
+
+func GetPrincipal(args *sparkArgs) string {
+	principal, contains := args.properties["spark.yarn.principal"]
+	if contains {
+		return principal
+	} else {
+		if args.kerberosPrincipal != "" {
```
elezar: This is equivalent to `return args.kerberosPrincipal`.

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+		return errors.New("cannot use a TGT-by-value and a TGT-by-secret at the same time")
+	}
+	return nil
+}
+
+func GetPrincipal(args *sparkArgs) string {
+	principal, contains := args.properties["spark.yarn.principal"]
+	if contains {
+		return principal
+	} else {
```
elezar: It may be more idiomatic to not include the code in the `else` block.

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+	if args.tgtSecretPath != "" && args.tgtSecretValue != "" {
+		return errors.New("cannot use a TGT-by-value and a TGT-by-secret at the same time")
+	}
+	return nil
+}
+
+func GetPrincipal(args *sparkArgs) string {
+	principal, contains := args.properties["spark.yarn.principal"]
+	if contains {
+		return principal
```
elezar: Should this not check for a non-empty principal?

cli/dcos-spark/security.go
```diff
@@ -0,0 +1,236 @@
...
+	krb5confBase64, errBase64Blob := getStringFromTree(marathonJson, []string{"app", "env", sparkKrb5ConfBase64BlobKey})
+	kdcHostname, errHostname := getStringFromTree(marathonJson, []string{"app", "env", sparkKdcHostnameKey})
+	kdcPort, errPort := getStringFromTree(marathonJson, []string{"app", "env", sparkKdcPortKey})
+	kdcRealm, errRealm := getStringFromTree(marathonJson, []string{"app", "env", sparkKerberosRealmKey})
+
+	// We allow the user to set SPARK_SECURITY_KERBEROS_KDC_HOSTNAME and SPARK_SECURITY_KERBEROS_KDC_PORT, and
+	// SPARK_SECURITY_KERBEROS_REALM, these values will be used to template a krb5.conf. If the user sets
+	// SPARK_MESOS_KRB5_CONF_BASE64 it will be overwritten, but log a warning to be sure.
+	kdcPropCount := 0
+	if propertyGiven(kdcHostname, errHostname) {
```
elezar: The combination of `getStringFromTree` and `if propertyGiven()` block and be combined into a function.

conf/spark-env.sh
```diff
@@ -44,6 +44,15 @@ export MESOS_AUTHENTICATEE="com_mesosphere_dcos_ClassicRPCAuthenticatee"
 
 echo "spark-env: User: $(whoami)" >&2
 
+echo "PWD $(pwd)"
```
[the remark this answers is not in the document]
ArtRand: Sorry, this was my own debugging code that I forgot to remove.

conf/spark-env.sh
```diff
@@ -44,6 +44,15 @@ export MESOS_AUTHENTICATEE="com_mesosphere_dcos_ClassicRPCAuthenticatee"
...
+
+if [ -n "${SPARK_SECURITY_KERBEROS_KDC_HOSTNAME}" ] && [ -n "${SPARK_SECURITY_KERBEROS_KDC_PORT}" ] && [ -n "${SPARK_SECURITY_KERBEROS_REALM}" ]; then
+    echo "templating krb5.conf"
+    # working dir is /mnt/mesos/sandbox
+    CONFIG_TEMPLATE_KRB5CONF=../../../etc/krb5.conf.mustache,../../../etc/krb5.conf $BOOTSTRAP -template -resolve=false --print-env=false -install-certs=false
```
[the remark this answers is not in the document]
ArtRand: Unfortunately, `bootstrap` will not use absolute paths.

docker/Dockerfile
```diff
@@ -76,6 +76,7 @@ RUN ln -s /var/lib/runit/service/spark /etc/service/spark
 RUN ln -s /var/lib/runit/service/nginx /etc/service/nginx
 
 ADD dist /opt/spark/dist
+ADD krb5.conf.mustache /etc/
```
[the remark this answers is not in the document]
ArtRand: Yes, partly. Spark runs a container, so everything needs to be internal to the container at start time.

docker/krb5.conf.mustache
```diff
@@ -0,0 +1,7 @@
+[libdefaults]
+default_realm = {{SPARK_SECURITY_KERBEROS_REALM}}
```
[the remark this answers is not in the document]
ArtRand: Environment variables with the `SPARK_` and `MESOS_` prefix are forwarded from the current environment when using the Dispatcher client API(https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/deploy/rest/RestSubmissionClient.scala#L404). Tl;dr having the `SPARK_` prefix means that the image will forward the environment variables without using the go cli.

docs/kerberos.md
```diff
@@ -96,14 +126,20 @@ Submit the job with the ticket:
...
 # Kafka Kerberos
 
-Spark can consume data from a Kerberos-enabled Kafka cluster. Connecting Spark to secure Kafka does not require special installation parameters, however does require the Spark Driver _and_ the Spark Executors can access the following files:
+Spark can consume data from a Kerberos-enabled Kafka cluster. Connecting Spark to secure Kafka does not require special
+installation parameters, however does require the Spark Driver _and_ the Spark Executors can access the following files:
 
-*   Client JAAS (Java Authentication and Authorization Service) file. This is provided using Mesos URIS with `--conf spark.mesos.uris=<location_of_jaas>`.
+*   Client JAAS (Java Authentication and Authorization Service) file. This is provided using Mesos URIS with `--conf
+    spark.mesos.uris=<location_of_jaas>`.
 *   `krb5.conf` for your Kerberos setup. Similarly to HDFS, this is provided using a base64 encoding of the file.
```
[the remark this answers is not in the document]
ArtRand: It's _potentially_ the same.

universe/config.json
```diff
@@ -70,57 +70,38 @@
...
                     "type": "object",
-                    "properties": {
-                        "krb5conf": {
-                            "description": "Base64 encoded krb5.conf file.  Copied to the drivers and executors so that they may access your KDC.  Also used by the history server.",
-                            "type": "string",
-                            "default": ""
-                        }
-                    }
-                },
-                "ssl": {
```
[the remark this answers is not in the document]
ArtRand: Nope, these were old from when we used the Spark binary directly on your machine.

universe/marathon.json.mustache
```diff
@@ -5,21 +5,21 @@
...
     "user": "{{service.user}}",
     "env": {
-{{#security.kerberos.krb5conf}}
-        "SPARK_MESOS_KRB5_CONF_BASE64": "{{security.kerberos.krb5conf}}",
-{{/security.kerberos.krb5conf}}
+{{#security.kerberos.enabled}}
+        {{#security.kerberos.krb5conf}}
+            "SPARK_MESOS_KRB5_CONF_BASE64": "{{security.kerberos.krb5conf}}",
+        {{/security.kerberos.krb5conf}}
+        {{#security.kerberos.kdc.hostname}}
```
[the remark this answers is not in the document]
ArtRand: See above.

cli/dcos-spark/security.go
ArtRand: `ValidateInputs` should make sure these are never empty.

cli/dcos-spark/security.go
ArtRand: `appendToProperty` and `appendToPropertyIfSet` already handle comma-separated strings, although I agree this is a little rough, it works for here. Ideally, we'd have a more intuitive API for passing secrets to Spark

cli/dcos-spark/security.go
ArtRand: No, the they are present as a placeholder. The way secrets work is they are "zipped" together. I made sure to add the environment variable placeholder anytime we adding secrets.

cli/dcos-spark/security.go
ArtRand: I put all the checks in one place.

cli/dcos-spark/security.go
ArtRand: We do. In Spark we don't (currently) automate the generation of the TLS artifacts. Therefore we allow the user to specify whichever they would like, but `TLS` as the default.

cli/dcos-spark/security.go
ArtRand: If the principal is empty then we won't start down the kerberos fork:
```go
// kerberos configuration:
	kerberosPrincipal := GetPrincipal(args)
	if kerberosPrincipal != "" {
		err := ValidateInputs(args)
		if err != nil {
			return "", err
		}
```

tests/utils.py
```diff
@@ -148,7 +148,12 @@ def _get_spark_options(use_hdfs, use_history, marathon_group, strict_mode, user)
         options["hdfs"]["config-url"] = "http://api.hdfs.marathon.l4lb.thisdcos.directory/v1/endpoints"
         options["security"] = options.get("security", {})
         options["security"]["kerberos"] = options["security"].get("kerberos", {})
-        options["security"]["kerberos"]["krb5conf"] = HDFS_KRB5_CONF
+        options["security"]["kerberos"]["enabled"] = True
+        options["security"]["kerberos"]["realm"] = "LOCAL"
+        options["security"]["kerberos"]["kdc"] = options["security"]["kerberos"].get("kdc", {})
+        options["security"]["kerberos"]["kdc"]["hostname"] = "kdc.marathon.autoip.dcos.thisdcos.directory"
+        options["security"]["kerberos"]["kdc"]["port"] = 2500
+        #options["security"]["kerberos"]["krb5conf"] = HDFS_KRB5_CONF
```
susanxhuynh: This line can be removed.

docs/kerberos.md
```diff
@@ -10,50 +10,78 @@ menuWeight: 120
...
 
 ## Spark Installation
 
-Spark (and all Kerberos-enabed) components need a valid `krb5.conf` file. You can setup the Spark service to use a single `krb5.conf` file for all of the its drivers.
+Spark (and all Kerberos-enabed) components need a valid `krb5.conf` file. You can setup the Spark service to use a
+single `krb5.conf` file for all of the its drivers.
 
 1.  A `krb5.conf` file tells Spark how to connect to your KDC.  Base64 encode this file:
 
         cat krb5.conf | base64 -w 0
```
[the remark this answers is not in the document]
ArtRand: it's not?

docs/kerberos.md
```diff
@@ -126,4 +162,6 @@ Finally, you'll likely need to tell Spark to use the JAAS file:
         --conf spark.executor.extraJavaOptions=-Djava.security.auth.login.config=/mnt/mesos/sandbox/<jaas_file>
```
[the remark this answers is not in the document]
ArtRand: Looking into this.

docs/walkthroughs/installing-secure-spark.md
```diff
@@ -25,6 +25,7 @@
       },
       "security": {
           "kerberos": {
+              "enabled": true,
```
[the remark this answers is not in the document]
ArtRand: I don't think we can/should fully deprecate it, it's still the easiest way to have any custom settings you'd like.

conf/spark-env.sh
```diff
@@ -44,6 +44,13 @@ export MESOS_AUTHENTICATEE="com_mesosphere_dcos_ClassicRPCAuthenticatee"
 
 echo "spark-env: User: $(whoami)" >&2
 
+if [ -n "${SPARK_SECURITY_KERBEROS_KDC_HOSTNAME}" ] && [ -n "${SPARK_SECURITY_KERBEROS_KDC_PORT}" ] && [ -n "${SPARK_SECURITY_KERBEROS_REALM}" ]; then
+    echo "Templating krb5.conf from environment"
```
ArtRand: this and other logging should go to `stderr`

docs/kerberos.md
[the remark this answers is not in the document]
ArtRand: Depending on the decision on how we decide to expose the user to provide a JAAS file I'll update these instructions.

ArtRand:
Thanks for the review @susanxhuynh and @elezar, I believe I've addressed the majority of the concerns. PTAL when you('all) have a second. Thanks!

susanxhuynh:
Nice refactor, thanks

docs/kerberos.md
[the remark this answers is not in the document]
elezar: I'm not sure. I'm used to specifying it explicitly `'''bash` (where `'` here implies a backtick).

elezar:
Thanks @art. Looks good.

PR #258, opened 2018-01-23
[TESTS][SPARK-604] Made the driver role and service account configurable

susanxhuynh:
Changes to tests to allow setting quota and running in strict mode.
Updated dcoscli version to align with dcos-commons.


ArtRand:
For the entirety of the 2-quota system, the Dispatcher needs to register under a role with Quota resources. I don't see that change here? Is it intentionally omitted?

ArtRand:
LGTM

PR #260, opened 2018-01-25
[SPARK-638] update install docs for strict mode

ArtRand:


docs/troubleshooting.md
```diff
@@ -9,19 +9,56 @@ menuWeight: 125
...
+        Caused by: java.net.UnknownHostException: ip-172-31-4-148: Name or service not known
+            at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
+            at java.net.InetAddress$2.lookupAllHostAddr(InetAddress.java:928)
+            at java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1323)
+            at java.net.InetAddress.getLocalHost(InetAddress.java:1500)
+            ... 10 more
+    18/01/25 17:42:57 INFO ShutdownHookManager: Shutdown hook called
+    ```
+
+    In this case, remove the environment variable called `NO_BOOTSTRAP` from the Dispatcher config, either via the UI, editing the task or set to false in the options.json, and restart the
```
susanxhuynh: I don't think you can set this in the options.json, at least when installing. Would it be through an "update" command?

ArtRand:
@susanxhuynh RE `dcos-docs-site` I guess so, we could back port the docs. The ones in this PR should be generic enough for all of the recent (back to <2.X.X-X.X.X) releases.

PR #266, opened 2018-01-30
add edit window for history server and remove dead bash code

ArtRand:


docker/runit/init.sh
```diff
@@ -32,44 +32,5 @@ sed -i "s,<PROTOCOL>,${SPARK_SSL_PROTOCOL}," /etc/nginx/conf.d/spark.conf
...
-	# export cert and key from p12
-	openssl pkcs12 -nokeys -passin pass:"${SPARK_SSL_KEYSTOREPASSWORD}" -in "$KEYDIR/spark.p12" -out /etc/nginx/spark.crt
-	openssl pkcs12 -nocerts -nodes -passin pass:"${SPARK_SSL_KEYSTOREPASSWORD}" -in "$KEYDIR/spark.p12" -out /etc/nginx/spark.key
-	chmod 600 /etc/nginx/spark.{crt,key}
-
-	rm -rf "$KEYDIR"
-fi
-
-# Move hadoop config files, as specified by hdfs.config-url, into place.
-if [[ -f hdfs-site.xml && -f core-site.xml ]]; then
```
elezar: Are these no longer required since we pull them from HDFS directly?

docker/runit/init.sh
```diff
@@ -32,44 +32,5 @@ sed -i "s,<PROTOCOL>,${SPARK_SSL_PROTOCOL}," /etc/nginx/conf.d/spark.conf
...
-    mkdir -p "${HADOOP_CONF_DIR}"
-    cp hdfs-site.xml "${HADOOP_CONF_DIR}"
-    cp core-site.xml "${HADOOP_CONF_DIR}"
-fi
-
-# Move kerberos config file, as specified by security.kerberos.krb5conf, into place.
-# this only affects the krb5.conf file for the dispatcher
-if [[ -n "${SPARK_MESOS_KRB5_CONF_BASE64}" ]]; then
```
elezar: This is now all handled in `spark-env.sh` right?

docker/runit/init.sh
```diff
@@ -32,44 +32,5 @@ sed -i "s,<PROTOCOL>,${SPARK_SSL_PROTOCOL}," /etc/nginx/conf.d/spark.conf
 # bytes cipher strings of Java.
 # sed -i "s,<ENABLED_ALGORITHMS>,${SPARK_SSL_ENABLEDALGORITHMS//,/:}," /etc/nginx/conf.d/spark.conf
 
-# extract cert and key from keystore, write to /etc/nginx/spark.{crt,key}
-if [[ "${SPARK_SSL_ENABLED}" == true ]]; then
```
elezar: This is no longer required because we use a similar method as the other SDK frameworks? Does this mean that the DC/OS secret store is required for SSL to work?

susanxhuynh: @elezar This was used by the CLI before we moved to the Go CLI. Now our CLI uses the github.com/mesosphere/dcos-commons/cli/client library (same as SDK frameworks) to create an SSL connection with a cluster.

docker/runit/init.sh
susanxhuynh: This code is now executed in conf/spark-env.sh: https://github.com/mesosphere/spark-build/blob/master/conf/spark-env.sh#L11

docker/runit/init.sh
susanxhuynh: Yes

PR #276, opened 2018-02-17
[TESTS][SPARK-659] Updated HDFS.auth_to_local, changed the principal to "alice"

susanxhuynh:
* Updated HDFS config
  * New kerberos config schema
  * New auth_to_local setting (copied from HDFS test)
* When running a job:
  * Changed the principal to “alice@REALM”
  * Added `--conf driverEnv.SPARK_USER`
* chmod the history server directory to make it writable by all users
* Updated docs


docs/kerberos.md
```diff
@@ -146,13 +151,25 @@ You can also set the base64 encoded krb5.conf after install time:
 **Note** This setting `SPARK_MESOS_KRB5_CONF_BASE64` will overwrite/override any settings set with
 `SPARK_SECURITY_KERBEROS_KDC_HOSTNAME`, `SPARK_SECURITY_KERBEROS_KDC_PORT`, and `SPARK_SECURITY_KERBEROS_REALM`
 
+### Setting the Spark User
+
+By default, when Kerberos is enabled, Spark runs as the OS user coresponding to the
```
elezar: Spelling: "coresponding" => "corresponding"

docs/kerberos.md
```diff
@@ -146,13 +151,25 @@ You can also set the base64 encoded krb5.conf after install time:
...
+Kerberos principal. For example, the principal "alice@LOCAL" would likely translate to "alice".
```
elezar: I would say that we should use "corresponding to the primary of the specified Kerberos principal."

elezar: Suggestion: "For example, the principal "alice@LOCAL" would ~likely~ map to the username "alice"."
(The only edit I feel strongly about here is dropping "likely").

docs/kerberos.md
```diff
@@ -146,13 +151,25 @@ You can also set the base64 encoded krb5.conf after install time:
...
+Since "alice" would probably not be available as an OS user, it is recommended to
```
elezar: Suggestion: "If it is known that "alice" is not available as an OS user, the Spark users should be specified as "root" or "nobody" instead:"

At the very least, I would not use the work "probably".

tests/test_hdfs.py
```diff
@@ -14,18 +14,21 @@
...
 
+from tests import hdfs_auth
 from tests import utils
 
 
 log = logging.getLogger(__name__)
 DEFAULT_HDFS_TASK_COUNT=10
 GENERIC_HDFS_USER_PRINCIPAL = "hdfs@{realm}".format(realm=sdk_auth.REALM)
+ALICE_PRINCIPAL = "alice@{realm}".format(realm=sdk_auth.REALM)
 KEYTAB_SECRET_PATH = os.getenv("KEYTAB_SECRET_PATH", "__dcos_base64___keytab")
```
elezar: Should this be an environment variable?

tests/test_hdfs.py
```diff
@@ -64,21 +67,27 @@ def hdfs_with_kerberos(configure_security_hdfs):
                 )
             )
         principals.append(GENERIC_HDFS_USER_PRINCIPAL)
+        principals.append(ALICE_PRINCIPAL)
```
elezar: See https://github.com/mesosphere/dcos-commons/blob/master/frameworks/hdfs/tests/test_kerberos_auth.py#L47 for a more consistent way of handling Kerberos principals and setting up the kerberos environment.

tests/test_hdfs.py
```diff
@@ -116,16 +125,23 @@ def setup_hdfs_client(hdfs_with_kerberos):
...
 
+def hdfs_cmd(cmd):
+    sdk_tasks.task_exec(HDFS_CLIENT_ID, "bin/hdfs dfs -{}".format(cmd))
+
+
 @pytest.fixture(scope='module')
 def setup_history_server(hdfs_with_kerberos, setup_hdfs_client, configure_universe):
     try:
-        sdk_tasks.task_exec(HDFS_CLIENT_ID, "bin/hdfs dfs -mkdir /history")
+        hdfs_cmd("mkdir /history")
```
elezar: Should we re-run `kinit` as the hdfs super user here?

tests/test_hdfs.py
```diff
@@ -166,21 +182,26 @@ def setup_spark(hdfs_with_kerberos, setup_history_server, configure_security_spa
 def test_terasort_suite():
     jar_url = 'https://downloads.mesosphere.io/spark/examples/spark-terasort-1.1-jar-with-dependencies_2.11.jar'
 
+    data_dir = "hdfs:///users/alice"
+    terasort_in = "{}/{}".format(data_dir, "terasort_in")
+    terasort_out = "{}/{}".format(data_dir, "terasort_out")
+    terasort_validate = "{}/{}".format(data_dir, "terasort_validate")
+
     teragen_args=["--class", "com.github.ehiggs.spark.terasort.TeraGen"] + KERBEROS_ARGS
```
elezar: Instead of adding the `KERBEROS_ARGS` for each of the steps, we could wrap `utils.run_tests` as a local function and append the arguments there before calling `utils.run_tests`.

tests/test_hdfs.py
```diff
@@ -166,21 +182,26 @@ def setup_spark(hdfs_with_kerberos, setup_history_server, configure_security_spa
...
     utils.run_tests(app_url=jar_url,
-                    app_args="1g hdfs:///terasort_in",
+                    app_args="1g {}".format(terasort_in),
```
elezar: If we implement a wrapper, we could also accept a `list` for the `app_args` and then pass `" ".join(app_args)` to `utils.run_tests()`.

tests/test_hdfs.py
```diff
@@ -209,7 +230,7 @@ def has_running_executors():
                 "--conf", "spark.executors.cores=4"]
 
     driver_id = utils.submit_job(app_url=utils.SPARK_EXAMPLES,
-                                 app_args="10.0.0.1 9090 hdfs:///netcheck hdfs:///outfile",
+                                 app_args="10.0.0.1 9090 hdfs:///users/alice/netcheck hdfs:///users/alice/outfile",
```
elezar: Was there a motivation for not using the same `data_dir` variable here?

docs/kerberos.md
skonto: That is the default behavior for kerberos to hdfs translation btw.

docs/kerberos.md
skonto: We need to explain why there is this inconsistency.  Also it is possible to have an OS user eg. core on CoreOS and not on the executor image. The same issue appears.

tests/test_hdfs.py
elezar: Sounds good. Let's address this in a follow-up ticket then.

tests/test_hdfs.py
```diff
@@ -229,8 +231,9 @@ def has_running_executors():
                 "--conf", "spark.cores.max=8",
                 "--conf", "spark.executors.cores=4"]
 
+    data_dir = "hdfs:///users/alice"
     driver_id = utils.submit_job(app_url=utils.SPARK_EXAMPLES,
-                                 app_args="10.0.0.1 9090 hdfs:///users/alice/netcheck hdfs:///users/alice/outfile",
+                                 app_args="10.0.0.1 9090 {dir}/netcheck {dir}/outfile".format(dir=data_dir),
```
elezar: Out of scope of this PR, but we could use the same (or similar) wrapper `_run_terasort_job` here.

PR #282, opened 2018-03-06
[DOCS][SPARK-620] Update docs for binary secrets in DC/OS 1.11

susanxhuynh:
This PR updates the Spark docs regarding the support for binary secrets in DC/OS 1.11.

- In the section on Binary Secrets, I have incorporated the changes that were made in dcos-commons (https://github.com/mesosphere/dcos-commons/pull/2213)
- In the docs pages, I have removed the `__dcos_commons__` prefix from examples, pointing 1.10 users to the above section to read about encoding.
- In the Walkthroughs, which are not published in docs (you can only read them in Github), I have left the examples as is, with a note that encoding is not necessary in 1.11.


docs/security.md
```diff
@@ -36,24 +36,46 @@ to the Spark Dispatcher instance.
...
+Now that the file is encoded it can be stored as a secret.
 
 ```bash
 $ dcos security secrets  create -f kerb5.keytab.base64-encoded  some/path/__dcos_base64__mysecret
 ```
-When you reference the `__dcos_base64__mysecret` secret in your service, the content of the secret will be first base64-decoded, and then copied and made available to your Spark application. Refer to a binary secret only as a file such that it will be automatically decoded and made available as a temporary in-memory file mounted within your container (file-based secrets).
+
+**Note:** The secret name **must** be prefixed with `__dcos_base64__`.
+
+When the `some/path/__dcos_base64__mysecret` secret is referenced in your `dcos spark run` command, its base64-decoded
```
elezar: Question: Is it referenced in the command or in some service specification (as is used here: https://github.com/mesosphere/dcos-commons/blob/6da53edd1ff5392986bef5096a8c7a96470c64eb/docs/pages/_includes/services/overview.md#binary-secrets)? I'm ok leaving it as is though.

docs/security.md
```diff
@@ -103,7 +125,7 @@ sources (i.e. files and environment variables). For example
 ```
 will place the content of `spark/my-secret-file` into the `PLACEHOLDER` environment variable and the `target-secret-file` file
 as well as the content of `spark/my-secret-envvar` into the `SECRET_ENVVAR` and `placeholder-file`. In the case of binary
-secrets (tagged with `__dcos_base64__`, for example) the environment variable will still be empty because environment
+secrets, the environment variable will still be empty because environment
```
elezar: Nit: I don't think this comma is required.

elezar: The wording "environment variables cannot be assigned to binary values" seems in the wrong order (or the "to" should be removed).

docs/security.md
```diff
@@ -138,24 +160,16 @@ The keystore and truststore are created using the [Java keytool][12]. The keysto
...
-
-```bash
-cat keystore | base64 -w 0 > keystore.base64
-cat keystore.base64
-/u3+7QAAAAIAAAACAAAAAgA...
-```
-
-**Note:** The base64 string of the keystore will probably be much longer than the snippet above, spanning 50 lines or
-so.
+**DC/OS 1.10 or lower:** Since both stores are binary files, they must be base64-encoded before being placed in the
```
elezar: The DC/OS 1.10 note could be placed *after* the instructions?

docs/security.md
susanxhuynh: In Spark, it's referenced in the Spark "run" command.

docs/security.md
susanxhuynh: I will fix it.

PR #285, opened 2018-03-30
[SPARK-688] Upgraded libmesos version and added test, test_task_not_lost()

susanxhuynh:
Before my PR:
- Every Spark app would get a TASK_LOST on Executors as soon as the driver starts up.

Changes in this PR:
- Adds a test, test_task_not_lost(), that checks if a TASK_LOST occurs upon driver start up.
- Upgrades the libmesos to a version that has a patch for this bug, MESOS-8171.

Additional changes:
- Minor fixes in the Makefile for development workflow.
- Added support for AWS session token for local testing.


susanxhuynh:
Verified that test_task_not_lost() fails, without the libmesos fix.

susanxhuynh:
@samvantran Please review and let me know if you have any questions.

susanxhuynh:
Yeah, it would be good to automate the dev workflow (as opposed to the CI workflow). Maybe we can borrow some ideas from the dcos-commons repo.

PR #288, opened 2018-04-18
[SPARK-692] Minimum GPU resources for Spark app (spark.mesos.executor.gpus)

susanxhuynh:
Added some tests of a new config property `spark.mesos.executor.gpus`. The tests are disabled by default because they require a GPU cluster. They pass when Spark is built with this PR: https://github.com/apache/spark/pull/21033

Tests added
- test_executor_gpus_allocated: use `spark.mesos.executor.gpus` to guaranteed a specific number of gpus per executor.
- test_executor_gpus_exceeds_available_gpus: if the requested `spark.mesos.executor.gpus` is more than available, the app should not run.
- test_gpus_max: test of `spark.mesos.gpus.max`.


susanxhuynh:
I need to rebase this PR.

PR #292, opened 2018-04-25
configuration: normalize the config.json properties to be snake case

hectorj2f:
This PR aims to normalize the config.json properties to be snake_case, as the other properties. There is a mix right now.

This PR needs to be discussed with PRODUCT, it'll imply to change the existing `config.json` files.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,6 +1,6 @@
 ---
 layout: layout.pug
-navigationTitle: 
+navigationTitle:
 excerpt:
 title: Install and Customize
 menuWeight: 0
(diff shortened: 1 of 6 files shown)


akirillov:
While standardising all the properties to the snake-case is a great idea, it is a breaking change for existing installations: all the users have to update their options.json to match the new property names.

as there's no activity on this plus conflicts, let's close this PR?

cc @samvantran

akirillov:
Closing this due to no activity, conflicts, and the breaking nature of the change.

PR #293, opened 2018-04-26
[SPARK-717] monitoring: add a spark prometheus dashboard

hectorj2f:
I thought we might want to add the dashboard to monitor some properties exposed by Spark when running on DC/OS. 

To use it, you just need to import the dashboard in Grafana.


hectorj2f:
@susanxhuynh Do we want to have these dashboards as part of the repo as done in `dcos-jenkins-service`?

hectorj2f:
@samvantran could we also add the GPU dashboard ?

PR #295, opened 2018-04-27
Update README make dev-dist step

samvantran:
I ran into issues when trying to build on a Mac. Looks like Macs have outdated `make` binaries. 
Creating the docker image and building from inside that resolves the issue.


samvantran:
Closing this in favor of #297

PR #297, opened 2018-05-01
SPARK-727 Resync test/build tooling from current dcos-commons

nickbp:
- Bump `testing/` and `tools/` to latest dcos-commons
- De-duplicate functionality in current integration tests with calls that were more recently added to `testing/` (should hopefully resolve flakes)
- De-duplicate build functionality from current Makefiles to instead use tooling in `tools/` (should reduce need for custom build tooling)
- Always the cluster/stub-universe specified by the user via the environment, rather than requiring the user to manually remove old `stub-universe-url`/`cluster-url` files to rebuild/reconfigure them.
- For S3 testing, create/use random subdirs within `autodelete7d/` rather than using the root to avoid potential for collisions between simultaneous test runs.
- Switch test environment to the `dcos-commons` image used elsewhere, and also use the dcos-commons `test.sh`. As a result of this, `spark_s3.py` is switched from the deprecated `boto` library to `botocore` since only the latter is available in the image.


tools/Dockerfile
```diff
@@ -0,0 +1,5 @@
+# Dockerfile for the KDC image
```
elezar: out of scope: I'm pretty sure this should be removed from the upstream ...

tools/ci/launch_cluster.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
elezar: We could call this from the makefile?

tools/create_service_account.sh
```diff
@@ -5,16 +5,8 @@ MODE=
...
-    # Set some cluster configs if they are passed in.
-    echo At least 2 arguments, must be dcos_url and acs_token
-    DCOS_URL=$1
-    ACS_TOKEN=$2
-
-    dcos config set core.dcos_url $DCOS_URL
-    dcos config set core.dcos_acs_token $ACS_TOKEN
-    dcos config set core.ssl_verify false
-fi
+SERVICE_ACCOUNT_NAME=${1:-service-acct}
```
elezar: It is my understanding that we deprecated these scripts in the mono-repo.

tools/distribution/UPDATING.md
```diff
@@ -0,0 +1,157 @@
+# Updating this repository
```
elezar: It may not be 100% applicable to Spark, but this doc describes updating the `tool` and `testing` folders in addition to the SDK version.

tests/jobs/json/hdfs-delete-terasort-files.json
```diff
@@ -1,17 +0,0 @@
-{
```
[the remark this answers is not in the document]
nickbp: The content of these files was moved into the tests that were using them. They had some templating and this made that much more straightforward. This also mirrors how cassandra's tests use metronome jobs.

tools/Dockerfile
nickbp: Agreed, just a blind copy here

tools/ci/launch_cluster.sh
nickbp: Good point. Looks like `test.sh` calls `test_runner.sh`, which then calls this if `CLUSTER_URL` is unspecified. Switched to using this instead of having a lot of custom cluster launch logic in the spark Makefile.

Makefile
```diff
@@ -19,148 +19,102 @@ docker-build:
...
+		exit 1
+	fi
+	tar xvf $${SPARK_BUILDS} -C $(DIST_DIR)
+
 	rm -rf $(BUILD_DIR)/docker
 	mkdir -p $(BUILD_DIR)/docker/dist
 	cp -r $(DIST_DIR)/spark-*/. $(BUILD_DIR)/docker/dist
 	cp -r conf/* $(BUILD_DIR)/docker/dist/conf
 	cp -r docker/* $(BUILD_DIR)/docker
-	pushd $(BUILD_DIR)/docker; \
```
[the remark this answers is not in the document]
nickbp: Oh, didn't realize that there were per-platform issues with `make`. Would it be OK if I just converted this Makefile to a shell script to avoid that? It'd still be of the form `build.sh <command>`, and there can continue to be a Makefile "frontend" with today's commands so that people's muscle memory isn't too disrupted.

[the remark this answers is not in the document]
susanxhuynh: Moving to a shell script would be fine. The main dependency on the Makefile is the TeamCity job ... we'd have to reconfigure it if we changed that interface.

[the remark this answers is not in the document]
nickbp: Makes sense, ok I'll take a stab at it.

tests/jobs/json/hdfs-delete-terasort-files.json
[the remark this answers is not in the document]
elezar: Yes, I did find them later. Thanks.

PR #299, opened 2018-05-02
[SPARK-734] Add script to launch kafka-spark streaming jobs

samvantran:
Addresses: [SPARK-734](https://jira.mesosphere.com/browse/SPARK-734)

This PR adds a configurable script to launch a simple Kafka-Spark streaming word count app at scale. You specify the number of dispatchers, consumers, and desired total runtime. 
- Included a README


scale-tests/streaming_test.py
```diff
@@ -0,0 +1,101 @@
...
+    producer_config = ["--conf", "spark.cores.max=2", "--conf", "spark.executor.cores=2",
+                       "--class", "KafkaFeeder"] + common_args
+
+    producer_id = utils.submit_job(app_url=jar_uri,
+                                   app_args=producer_args,
+                                   app_name=spark_app_name,
+                                   args=producer_config,
+                                   verbose=False)
+
+    shakedown.wait_for(lambda: test_kafka._producer_launched(), ignore_exceptions=False, timeout_seconds=600)
```
susanxhuynh: If the timeout is reached, this will throw an exception and the script will fail. I think we can omit this line ... just submit and continue. Then if the job gets stuck in the Dispatcher queue because of some bug, we can measure how long it's stuck in the metrics we're monitoring (that Nick is adding) and which may be longer than 600 seconds.

scale-tests/streaming_test.py
```diff
@@ -0,0 +1,101 @@
...
+    consumer_args = " ".join([broker_dns, topic, stop_count, kerberos_flag])
+
+    for i in range(0, num_consumers):
+        utils.run_tests(app_url=jar_uri,
+                        app_args=consumer_args,
+                        expected_output="Read {} words".format(stop_count),
+                        app_name=spark_app_name,
+                        args=consumer_config)
+
+    utils.kill_driver(producer_id, spark_app_name)
```
susanxhuynh: I've noticed this line is a little flaky. Sometimes I see the producer doesn't get killed in the Soak cluster, it just runs forever. I don't know of a quick fix, except to modify the Producer app and add a "stop_count" parameter like in the Consumer. For now, I guess it won't be an issue because the Producer would get killed anyway when the dispatchers are uninstalled.

scale-tests/streaming_test.py
samvantran: okay I'll omit

scale-tests/streaming_test.py
samvantran: Decided to leave it out bc like you said, uninstalling dispatchers shuts down producers. 2 for 1 deal :)

scale-tests/streaming_test.py
```diff
@@ -0,0 +1,102 @@
...
+KERBEROS_FLAG = "false"
+JAR_URI = "https://s3-us-west-2.amazonaws.com/infinity-artifacts/soak/spark/dcos-spark-scala-tests-assembly-0.1-SNAPSHOT.jar"
+PRODUCER_AVG_NUM_WORDS_PER_MIN = 500
+
+
+def run_pipeline(num_consumers, desired_runtime, spark_app_name):
+    broker_dns = _kafka_broker_dns()
+    topic = "topic-{}".format(spark_app_name)
+
+    num_words_to_read = int(desired_runtime * PRODUCER_AVG_NUM_WORDS_PER_MIN / num_consumers)
```
susanxhuynh: I don't understand this. Don't all the consumers consume from the beginning offset of the topic? (They all start at offset 0 and read to the end of the topic.) So, the number of words shouldn't depend on the number of consumers.

scale-tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Streaming
+To run the Kafka-Spark Streaming word count app:
+- Install kafka
+  ```bash
+  dcos package install kafka --yes
+  ```
+- Deploy spark dispatchers
+  ```bash
+  Usage: deploy-dispatchers.py [num_dispatchers] [service_name_base] [output_file] (options_json_file)
```
susanxhuynh: This changed slightly since Scott's latest commit: https://github.com/mesosphere/spark-build/commit/2ac5e46822120562c247670fb362066801a41854

scale-tests/streaming_test.py
```diff
@@ -0,0 +1,102 @@
...
+                     verbose=False)
+
+
+if __name__ == "__main__":
+    usage = """
+        Setup: export PYTHONPATH=../spark-testing:../testing:../tests
+        Usage: python streaming_test.py [dispatcher_file] [num_consumers_per_producer] [desired_runtime_in_mins]
+    """
+
+    if len(sys.argv) < 4:
```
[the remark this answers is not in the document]
elezar: I see that @sschneid used `docopt`. Not sure if you have a preference?

scale-tests/streaming_test.py
samvantran: just to keep record since we discussed it already - it doesn't seem like consumers all read from the topic at the same time. Producers write events and only 1 consumer reads at a time. When it's done, the next consumer in line starts reading. That's why this function is dependent on `num_consumers`. 

I'll look into why this is and see if we can get all the consumers to read at the same time

scale-tests/streaming_test.py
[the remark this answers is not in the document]
samvantran: I don't and am happy to use either. This PR was quick and dirty so I agree we should use something to create named args. Since Scott already got underway with docopt I'll use that.

scale-tests/streaming_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+import sdk_cmd
+import spark_utils as utils
+
+KAFKA_PACKAGE_NAME = "kafka"
+KAFKA_SERVICE_NAME = "kafka"
+PRODUCER_SERVICE_NAME = "Spark->Kafka Producer"
+KERBEROS_FLAG = "false"
+JAR_URI = "https://s3-us-west-2.amazonaws.com/infinity-artifacts/soak/spark/dcos-spark-scala-tests-assembly-0.1-SNAPSHOT.jar"
+PRODUCER_AVG_NUM_WORDS_PER_MIN = 500
```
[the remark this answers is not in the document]
samvantran: This actually isn't configurable. It's something I observed from testing the app. The KafkaProducer seems to output about 8 words every 2 seconds so: `8 * 2 * 30 = 480`.

scale-tests/streaming_test.py
```diff
@@ -1,102 +1,115 @@
...
-    dispatchers_file = sys.argv[1]
-    print("dispatchers_file: {}".format(dispatchers_file))
-    with open(dispatchers_file) as f:
-        dispatchers = f.read().splitlines()
-    print("dispatchers: {}".format(dispatchers))
+    args = docopt(__doc__)
 
-    num_consumers = int(sys.argv[2])
-    print("num_consumers_per_producer: {}".format(num_consumers))
+    with open(args["<dispatcher_file>"]) as f:
```
elezar: Note for myself: I wasn't expecting the `<` and `>` to be needed.

scale-tests/streaming_test.py
```diff
@@ -1,102 +1,115 @@
...
 
-    desired_runtime = int(sys.argv[3])
-    print("desired_runtime_in_mins: {}".format(desired_runtime))
+    log.info("Installing kafka CLI...")
+    kafka_pkg_name = args["--kafka-package-name"]
+    sdk_cmd.run_raw_cli("package install {} --yes --cli".format(kafka_pkg_name))
 
-    print("Installing kafka CLI...")
-    rc, stdout, stderr = sdk_cmd.run_raw_cli("package install {} --yes --cli".format(KAFKA_PACKAGE_NAME))
+    jar_url = args["--jar"] or "https://s3-us-west-2.amazonaws.com/infinity-artifacts/soak/spark/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar"
```
elezar: Minor nit:  I think 
```python
jar_url = args["--jar"] if args["--jar"] else "https://s3-us-west-2.amazonaws.com/infinity-artifacts/soak/spark/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar"
```
may be more pythonic, but it's not critical.

scale-tests/streaming_test.py
samvantran: Seems docopt just takes the entire string as a param just like how the dashes are required for arg `["--jar"]`.

scale-tests/streaming_test.py
samvantran: works for me, i'll fix.

scale-tests/streaming_test.py
```diff
@@ -0,0 +1,115 @@
...
+        "--conf", "spark.port.maxRetries=32",
+        "--conf", "spark.scheduler.maxRegisteredResourcesWaitingTime=2400s",
+        "--conf", "spark.scheduler.minRegisteredResourcesRatio=1.0"
+    ]
+
+    broker_dns = _kafka_broker_dns()
+    num_words_to_read = int(desired_runtime * PRODUCER_NUM_WORDS_PER_MIN)
+
+    for dispatcher in dispatchers:
+        spark_app_name, driver_role = dispatcher.split(",")
```
sschneid: Should this be `spark_app_name, dispatcher_role, driver_role`, or should `dispatcher_role` be removed from the output of `deploy-dispatchers.py`? cc @susanxhuynh

samvantran: I actually had it with the `dispatcher role` in the middle but have seen different variations of whether to use it or not. This seemed to be the latest working version since the Marathon batch app points to a `dispatchers.out` file with only `dispatcher-name`, and `driver-role`.

susanxhuynh: @sam It should be okay now to include all three. The "deploy-dispatchers" script now generates all three in its output file.

samvantran: okay i'll add a placeholder for it since the script itself doesn't use the dispatcher-role

PR #303, opened 2018-05-04
[SPARK-736] Run scale tests from inside the cluster

susanxhuynh:
1. The batch test script runs in Marathon (spark-batch-workload.json.template) to address the fault injection concern: if the script remains stateless, it can be restarted by Marathon.

2. To address concerns about passwords, I've add a wrapper script (deploy-marathon-service.sh). It fills in the username / password template in "spark-batch-workload.json.template". You also pass in the test script args as the last argument to this deploy script (see below).

These would be the commands to start a test:
> $ dcos cluster setup https://{master_IP}
> $ dcos package repo add --index=0 spark-aws {stub_universe_url}
> $ python deploy-dispatchers.py 1 shdispatcher test.out --quota-dispatcher-cpus etc. ...
> $ aws s3 cp --acl public-read test.out s3://{bucket}/test.out
> $ ./deploy-marathon-service.sh spark-batch-workload.json.template bootstrapuser deleteme https://{bucket}.s3.amazonaws.com/test.out “test.out --submits-per-min 1”

Misc. Fixes
- Fixed the "quota list" call when there is no quota defined yet.
- Added "--supervise" and "failoverTimeout" to job config.
- Changed job duration to a random distribution that averages to about an hour.
- In uninstall-dispatchers.sh: fixed specifying input file.


scale-tests/spark-deploy-dispatchers.json
```diff
@@ -0,0 +1,37 @@
...
+      "QUOTA_DRIVER_CPUS": "4",
+      "QUOTA_DRIVER_MEM": "8192"
+    },
+    "placement": {
+      "constraints": []
+    },
+    "artifacts": [],
+    "maxLaunchDelay": 3600,
+    "docker": {
+      "image": "mesosphere/spark-build:latest"
```
elezar: Is this image valid, or is it overridden in the python files?

scale-tests/spark-batch-workload.json
```diff
@@ -0,0 +1,56 @@
...
+      "image": "mesosphere/spark-build:latest",
+      "forcePullImage": false,
+      "privileged": false,
+      "parameters": []
+    }
+  },
+  "cpus": 2,
+  "disk": 1024,
+  "env": {
+    "DCOS_PASSWORD": "deleteme",
```
elezar: Is there a different way to handle passwords?

scale-tests/scale_test.py
```diff
@@ -22,35 +22,43 @@
...
         app_url=MONTE_CARLO_APP_URL,
         app_args=app_args,
         verbose=False,
-        args=args)
+        args=args,
+        driver_role=driver_role)
 
 
 def submit_loop(launch_rate_per_min, dispatchers):
+    print("dispatchers: {}".format(dispatchers))
```
elezar: Is there a reason to use `print` instead of logging. I see that we're already using it elsewhere in the file, so this is probably out of scope for this PR.

scale-tests/scale_test.py
```diff
@@ -22,35 +22,43 @@
...
-            "--conf", "spark.mesos.executor.docker.image=mesosphere/spark-dev:931ca56273af913d103718376e2fbc04be7cbde0",
+            "--conf", "spark.mesos.executor.docker.image=mesosphere/spark-dev:f5dd540adffd9ab9e3e826e48d22e39ebc296567-aee120236ee37309892ddbee42acf0028096947f",
             # use Hector's image
-            "--conf", "spark.port.maxRetries=32"  # setting to allow up to 32 drivers on same node
+            "--conf", "spark.port.maxRetries=32",  # setting to allow up to 32 drivers on same node
             #"--conf", "spark.mesos.driverEnv.SPARK_USER=root", # Run as root on centos
+            "--supervise",
+            #"--conf", "spark.mesos.rejectOfferDuration=1000s",
+            "--conf", "spark.mesos.driver.failoverTimeout=30",
+            "--conf", "spark.mesos.task.labels=jpmc_test_id:spark_batch_run001"
```
elezar: Should we make the task label configurable? Is this code not also used by the streaming jobs?

scale-tests/spark-deploy-dispatchers.json
susanxhuynh: It's valid. It's the same image that is used in CI to run pytests. The ones that you may have seen in python files are images used by the Spark service itself (the image that runs Spark dispatchers and drivers).

scale-tests/spark-batch-workload.json
susanxhuynh: Metronome doesn't support secrets, but Marathon does. I could store the password in the secret store for the Marathon service.

scale-tests/scale_test.py
susanxhuynh: I'll replace it with logging.

scale-tests/scale_test.py
susanxhuynh: Yes, that makes sense. No, there's currently a separate script for streaming jobs. @samvantran looked into merging it but found it difficult. The big difference is that the streaming jobs get kicked off once and are long-running. The batch jobs get submitted continuously and end an hour after they're launched.

samvantran: I've copied over mostly every conf used in this script to the other one (didn't see this label one until now

susanxhuynh:
@elezar @samvantran I moved things around to address some of the comments: templated the username / password, added a wrapper script that does the templating, removed the env vars, removed the metronome config. Updated the PR description with the new workflow.

scale-tests/uninstall-dispatchers.sh
```diff
@@ -30,15 +31,15 @@ run_janitor () {
     dcos node ssh --leader --user=centos --master-proxy "$JANITOR_CMD"
```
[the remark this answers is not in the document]
mpereira: They do. And the `principal` seems to be `dcos-dev` instead of `spark-principal`.

PR #305, opened 2018-05-10
Use all kafka brokers for streaming_test.py + more configs

samvantran:
Resolves: [DCOS-33759](https://jira.mesosphere.com/browse/DCOS-33759)

This change uses all brokers retrieved from  cli command `dcos kafka endpoints broker`

Also adds additional configs:
- pass in custom docker images 
- set number of port retries since we will be scale testing on bigger 64-node machines

Example python command to run script:
```bash
python streaming_test.py dispatchers.out \
--desired-runtime 360 --num-consumers 20 --port-retries 64 --docker-image mesosphere/spark-dev:fea8e635e369596fb521e3aa140faecdbb0cacd0-5879a4ecf7d9a3781c70f3cdec64acff4f770e73
```


scale-tests/streaming_test.py
```diff
@@ -10,6 +10,8 @@
 
 Options:
     --desired-runtime <n>        desired time for consumers to run (in minutes) [default: 10]
+    --docker-image <image>       docker image to submit with spark job [default: mesosphere/spark:latest]
```
susanxhuynh: I think you can omit this config property altogether. I added this originally to try various images. But, now that we add a stub universe repo during testing, the dispatcher and drivers will automatically inherit whatever image is in that stub universe (without having to set spark.executor.docker.image explicitly).

samvantran: sounds good i'll take it out.. FYI it doesn't seem like this script will be used too often now that we have the more robust kafka-c* streaming app

samvantran: but this is good to have if we ever want to kick off a bunch of simple kafka jobs that don't require quota or a backend

elezar:
@samvantran since you addressed @susanxhuynh's comments regarding the docker image, I have merged this PR as I want to get the ones that are still around and dependent on this merged as well.

PR #312, opened 2018-05-21
Remove hardcoded executor docker image.

porridge:
Let spark inherit the dispatcher image.


elezar:
As per the discussion in #305 this looks good.

PR #314, opened 2018-05-21
[DCOS-37534] Make quotas optional and improve stdout masking.

porridge:
Trying to install a spark build that has no quota support spews scary messages, but should probably be OK for small-scale tests?


scale-tests/deploy-dispatchers.py
```diff
@@ -116,11 +115,13 @@ def deploy_dispatchers(
...
             dispatcher_role = "{}-dispatcher-role".format(service_name)
-            create_quota(name=dispatcher_role,
-                         cpus=quota_dispatcher_cpus, gpus=quota_dispatcher_gpus, mem=quota_dispatcher_mem)
+            if create_quotas:
+              create_quota(name=dispatcher_role,
+                           cpus=quota_dispatcher_cpus, gpus=quota_dispatcher_gpus, mem=quota_dispatcher_mem)
             driver_role = "{}-driver-role".format(service_name)
-            create_quota(name=driver_role,
-                         cpus=quota_driver_cpus, gpus=quota_driver_gpus, mem=quota_driver_mem)
+            if create_quotas:
```
samvantran: no need to check twice, `if create_quotas` is true on L118 you can create quota for both dispatchers and driver

scale-tests/deploy-dispatchers.py
```diff
@@ -116,11 +115,12 @@ def deploy_dispatchers(
 
             # create dispatcher & driver role quotas
             dispatcher_role = "{}-dispatcher-role".format(service_name)
-            create_quota(name=dispatcher_role,
-                         cpus=quota_dispatcher_cpus, gpus=quota_dispatcher_gpus, mem=quota_dispatcher_mem)
             driver_role = "{}-driver-role".format(service_name)
-            create_quota(name=driver_role,
-                         cpus=quota_driver_cpus, gpus=quota_driver_gpus, mem=quota_driver_mem)
+            if create_quotas:
+              create_quota(name=dispatcher_role,
```
mpereira: Is the indentation off here?

porridge:
Right, sorry, I'm used to 2-space indents. I'll fix it at the next
opportunity.
I guess I'm doomed to making mistakes like this if we have no pre-submit
linting.

porridge:
FTR, I bundled indent fix in https://github.com/mesosphere/spark-build/pull/320

PR #315, opened 2018-05-21
Update streaming README to use permanent spark jar

samvantran:
Best not to use a jar in the autodelete7d folder... so I moved it up a level


porridge:
I agree we should move it out of there, but I'm not sure if to the root directory.

I don't know how much junk is there already in the root directory of this bucket (`aws s3 ls s3://infinity-artifacts.s3.amazonaws.com/` says `NoSuchBucket` for me) but I'd put it under something like `/2018-cap-jpmc` to make cleanup easier in the future.

mpereira:
@porridge try `aws s3 ls s3://infinity-artifacts`.

porridge:
Thanks @mpereira that worked.
So there is just a handful of files in `/`, let's try to not add more.

elezar:
Yes, @samvantran could you move the permanent jars to a folder instead.

elezar:
@samvantran this was addressed in 03ec1ba8bfa2b09b682196db4ea616070b60c861, so I am going to go ahead and close this.

PR #318, opened 2018-05-22
Update GPU metric dashboard

samvantran:
Updating the GPU grafana dashboard with more stats on total GPUs free vs used, % of all GPUs used in the cluster, etc.

Here's a screenshot:

screen shot 2018-05-22 at 2 25 17 pm


samvantran:
The left graph is the sum of the right. There are 2 p2.8xlarge instances, so the left graph shows max 16 gpus ever available while right graph shows that at a per agent level, only 8 gpus can be used. I could stack the right graph and then both graphs would show 16 on the y-axis. Do you prefer that?

PR #319, opened 2018-05-22
[TESTS] Changed role names to "-drivers-role" and "-executors-role" in deploy-dispatchers script

susanxhuynh:
Changed the role names based on the tasks that consume the quota from those roles.
- "-drivers-role": quota is consumed by Driver tasks.
- "-executors-role": quota is consumed by Executor tasks.


susanxhuynh:
cc @samvantran @mpereira @porridge

porridge:
Or even do away with using the word dispatcher altogether, and call them:
`spark-NN-drivers-role` and `spark-NN-executors-role`. Where `spark-NN` is the Nth instance of spark package. Also note the `s`, because the quota/role covers all drivers and executors that are spark-NN's descendants.

susanxhuynh:
OK, I will go with the last suggestion: "-drivers-role" and "-executors-role". I will not change the prefix to "spark-NN" because the prefix is supplied by the user and is used to distinguish between different instances of spark running in the cluster.

PR #324, opened 2018-05-24
Fix setup_streaming.py to include right service_options

samvantran:
During Wednesday's test, I got runtime exceptions when running `kafka_cassandra_streaming_test.py`. I had to edit the infrastructure.json file with these changes:
```
{
  "kafka":{
    "kafka": {	<-- should be brokers
      "count":  <-- was expecting this to be filled in?
    },
  "cassandra": {
    "nodes": {
      "count": <-- was expecting this
    }
  }
}
```
This PR fixes up the `setup` script to include necessary key-values in the dicts in order to run `kafka_streaming_test.py` successfully. 

**Note:** this PR is a merge into PR #311 (because that's where I ran the instructions from)


scale-tests/setup_streaming.py
```diff
@@ -153,13 +153,16 @@ def get_expected_task_count(service_options: dict) -> int:
...
                                         "endpoint clientport", json=True)["dns"]
 
+        expected_count = get_expected_task_count
+
         service_options = {
-            "kafka": {
-                "kafka_zookeeper_uri": ",".join(zookeeper_dns)
+            "brokers": {
+                "kafka_zookeeper_uri": ",".join(zookeeper_dns),
+                "count": expected_count
```
porridge: I don't get this line. `get_expected_task_count` is a function, therefore `expected_count` is a function too. Why are we setting the value of `count` to a function here?

scale-tests/setup_streaming.py
```diff
@@ -153,13 +153,16 @@ def get_expected_task_count(service_options: dict) -> int:
                                         zookeeper_service["service"]["name"],
                                         "endpoint clientport", json=True)["dns"]
 
+        expected_count = get_expected_task_count
```
porridge: Why do we need the intermediate variable? We evaluate the function in each loop iteration, so there should not be any scope issues.

scale-tests/setup_streaming.py
```diff
@@ -181,8 +184,16 @@ def get_expected_task_count(service_options: dict) -> int:
     cassandra_config = args.get("--cassandra-config", "")
 
     services = []
+
+    expected_count = get_expected_task_count
```
porridge: Same questions here.

samvantran:
Sorry I didn't make the PR purpose clear earlier. During Wednesday's test, I got runtime exceptions related to these:
```
{
  "kafka":{
    "kafka": {	<-- should be brokers
      "count":  <-- was expecting this to be filled in?
    },
  "cassandra": {
    "nodes": {
      "count": <-- was expecting this
    }
  }
}
```
Specifically
- this line: https://github.com/mesosphere/spark-build/blob/kafka-spark-cassandra-streaming-fixes/scale-tests/kafka_cassandra_streaming_test.py#L189
and this line: https://github.com/mesosphere/spark-build/blob/kafka-spark-cassandra-streaming-fixes/scale-tests/kafka_cassandra_streaming_test.py#L192
would throw errors because those keys were not present in the `infrastructure.json` file created during the `setup_streaming.py` script.

so this PR was just adding a little more to the service config dicts that got merged so step 3 & 4 could execute properly.

scale-tests/setup_streaming.py
samvantran: @mpereira you're right, `count` should be under `brokers` separately: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/kafka/universe/config.json#L280

scale-tests/setup_streaming.py
samvantran: Sorry @porridge this should have been `expected_count = get_expected_task_count()` so the function was evaluated. I'll fix

scale-tests/setup_streaming.py
samvantran: same response above - i'll fix

elezar:
@porridge have your concerns now been addressed?

PR #327, opened 2018-05-25
Kafka spark cassandra streaming fixes edit elezar

elezar:
@samvantran here is a PR to correct the issue that you are addressing. Thanks for pointing these out.

Note the following:
* The dictionary accesses that were causing the runtime errors have been removed. These were being assigned to unused variables.
* I have now switched to using `dcos <package-name> describe` to get the service settings for the infrastructure file, instead of only returning user input.
* I also renamed a few local variables to make their intent clearer.

With regards to the use of the `get_expected_task_count` function. Note that this function accepts as input the `service` settings so as to determine whether the defaults have been overridden. This is the reason that it is not sufficient to set it once for all instances of the service. Furthermore, this function can only be called *after* any settings have been loaded from file and merged with the other user-specified settings.

Please let me know if there are any questions.


porridge:
Reviewed and merged.

PR #331, opened 2018-05-28
Fix package installation by not using shakedown.

porridge:
Looks like current version of dcos-commons docker image contains a
broken version of shakedown:

```
Traceback (most recent call last):
  File "./scale-tests/deploy-dispatchers.py", line 186, in <module>
    quota_executors_mem=arguments['--quota-executors-mem'])
  File "./scale-tests/deploy-dispatchers.py", line 137, in deploy_dispatchers
    options_json=options)
  File "/usr/local/lib/python3.5/dist-packages/shakedown/dcos/package.py", line 113, in install_package
    package_manager.install_app(pkg, options, service_name)```
TypeError: install_app() takes 3 positional arguments but 4 were given
```

This change switches to using sdk_install instead.


porridge:
@elezar is the broken shakedown something that we should track separately?

scale-tests/deploy-dispatchers.py
```diff
@@ -170,12 +166,15 @@ def deploy_dispatchers(
         }
     }
 
+    options_file = arguments['--options-json']
+    if options_file:
```
elezar: One note: We could add a bit more info here, but `json.load` will crash early.

elezar:
@porridge 
> @elezar is the broken shakedown something that we should track separately?

This should be a known issue, and I will try to find a JIRA ticket.

scale-tests/deploy-dispatchers.py
```diff
@@ -170,12 +166,15 @@ def deploy_dispatchers(
...
+        options = json.load(open(options_file, 'r'))
```
mpereira: Makes sense using `with open(...)` to close the file?

porridge: For a single open, in a short-running program, *and* read-only access I think putting it on one line is fine.

PR #332, opened 2018-05-28
Make default options clearer

elezar:
@porridge @mpereira minor changes for #331.


scale-tests/deploy-dispatchers.py
```diff
@@ -165,10 +164,23 @@ def deploy_dispatchers(
...
+    return options
+
+
+if __name__ == "__main__":
+    arguments = docopt(__doc__, version="deploy-dispatchers.py 0.0")
 
     options_file = arguments['--options-json']
     if options_file:
-        options = json.load(open(options_file, 'r'))
+        if not os.path.isfile(options_file):
```
porridge: IMHO such explicity check is not very useful, because it is superfluous (open does this check anyway), adds lines to the script which can contain bugs, and is very narrow (open can fail for many other reasons), hides potentially useful information (stack trace and exact error code/message).

PR #333, opened 2018-05-28
Add virtual networking to Kakfa / ZK / Cassandra settings

elezar:
This moves @porridge's settings from mesosphere/soak-cluster-configurations#303 into this repo.


scale-tests/configs/kafka-options.json
```diff
@@ -1,5 +1,10 @@
 {
+    "service": {
+        "virtual_network_enabled": true,
+        "virtual_network_name": "dcos"
+    },
     "brokers": {
-        "count": 5
+        "count": 5,
+        "disk_type": "ROOT"
```
mpereira: Is the `ROOT` volume change intentional?

elezar: Yes, it's to remind me that it needs to be changed to `MOUNT` (once we have the mount volumes).

mpereira:
@elezar can we merge this one?

PR #334, opened 2018-05-28
Don't crash spark_utils module loading.

mpereira:


spark-testing/spark_utils.py
```diff
@@ -17,24 +17,11 @@
 DCOS_SPARK_TEST_JAR_PATH = os.getenv(DCOS_SPARK_TEST_JAR_PATH_ENV, None)
 DCOS_SPARK_TEST_JAR_URL_ENV = "DCOS_SPARK_TEST_JAR_URL"
 DCOS_SPARK_TEST_JAR_URL = os.getenv(DCOS_SPARK_TEST_JAR_URL_ENV, None)
-if not DCOS_SPARK_TEST_JAR_URL and not os.path.exists(DCOS_SPARK_TEST_JAR_PATH):
```
elezar: should this be `or`?

mpereira: Looks like it.

mpereira: https://github.com/mesosphere/spark-build/pull/334/commits/f84786a385f8536c8cd2ca4d5b0b84bdd14cb2bf

spark-testing/spark_utils.py
```diff
@@ -175,6 +184,8 @@ def upload_mesos_test_jar():
 
 
 def dcos_test_jar_url():
+    _check_tests_assembly()
+
     if DCOS_SPARK_TEST_JAR_URL is None:
```
elezar: I know this is out of scope for this PR, but this looks a bit strange. We now have a module that *fails* if `DCOS_SPARK_TEST_JAR_URL` is not truthy, and then explicitly check it here.

OK, no, I take it back. Our condition checks that at least one of `DCOS_SPARK_TEST_JAR_URL` or `DCOS_SPARK_TEST_JAR_PATH ` is valid.

mpereira: I was confused there for a minute too.

spark-testing/spark_utils.py
elezar: No, it shouldn't be an `or` ... I was confused because these are different variables.

PR #335, opened 2018-05-28
Document the prometheus installation in README

elezar:
This adds a basic readme for deploying prometheus (as has been done on the current cluster).

It has a bit of a different tone to the other sections, so I didn't want to include it as a prerequisite.


scale-tests/README.md
```diff
@@ -350,3 +350,102 @@ export PYTHONPATH=../testing:../spark-testing
...
 generated by the set up step (#1).
+
+
+## Additional information
+
+### Prometheus
+
+#### (optional) Updating dcos-metrics
+
+On order (1.10.4 clusters), dcos-metrics needs to be updated on the agents.
```
porridge: Must be my bad English, but what does "On order" mean here?

PR #342, opened 2018-05-30
[TESTS][DCOS-37582] Support for strict mode in batch_test.py script

susanxhuynh:


scale-tests/batch_test.py
```diff
@@ -62,22 +63,33 @@ def _get_duration() -> int:
...
     app_args = "100000 {}".format(str(duration * 30))  # about 30 iterations per min.
 
-    spark_utils.submit_job(
-        service_name=dispatcher_name,
-        app_url=MONTE_CARLO_APP_URL,
-        app_args=app_args,
-        verbose=False,
-        args=config,
-        driver_role=driver_role)
+    if dispatcher["service"].get("service_account") is not None:  # only defined in strict mode
```
samvantran: There's an `sdk_utils` method for checking if `in_strict_mode` [here](https://github.com/mesosphere/spark-build/blob/master/testing/sdk_utils.py#L118) but it checks for an env var whereas you use the dispatcher file.

However, I would suggest the same approach of using a descriptive private method instead of using comments so the code is a little more readable

PR #343, opened 2018-05-30
[TESTS] Added two graphs to Spark Grafana dashboards

susanxhuynh:
(1) "Number of Job Failures by Dispatcher", 
(2) "Avg. Number of Queued Retry Jobs per Dispatcher"
queued-retries-2018-05-24
job-failures 2018-05-24


susanxhuynh:
cc @samvantran

susanxhuynh:
Yeah, there are two queues, one for retries ("pendingRetryDrivers"), the other for new jobs. Yeah, it has to wait for Mesos offers again. It looks like the retry jobs get priority over new jobs, when it comes to offers.

MesosClusterScheduler.scala
> // We first schedule all the supervised drivers that are ready to retry.
      // This list will be empty if none of the drivers are marked as supervise.

PR #345, opened 2018-05-31
Extend batch_test.py for GPU workloads

samvantran:
This extends the batch script to include launching GPU jobs in the same was as the monte carlo jobs.

There's a basic check to see if there are any gpus assigned and if so adds all the necessary confs for launching the jobs.


scale-tests/batch_test.py
```diff
@@ -123,10 +128,28 @@ def submit_loop(submits_per_min: int, dispatchers: typing.List[typing.Dict], use
...
+        app_url = MONTE_CARLO_APP_URL
+
     if args["--spark-mesos-driver-labels"] is not None:
         user_conf += ["--conf", "spark.mesos.driver.labels={}".format(args["--spark-mesos-driver-labels"])]
 
     if not args["--no-supervise"]:
         user_conf += ["--supervise"]
 
-    submit_loop(int(args["--submits-per-min"]), dispatchers, user_conf)
+    if args["--max-num-dispatchers"]:
```
elezar: Let's move this up to where the dispatchers are first loaded so as to keep all dispatcher-related logic together. We could even move it to a function so as to not make the intent clearer.

scale-tests/batch_test.py
```diff
@@ -123,10 +128,28 @@ def submit_loop(submits_per_min: int, dispatchers: typing.List[typing.Dict], use
                  "--conf", "spark.mesos.driver.failoverTimeout={}".format(args["--spark-mesos-driver-failover-timeout"])
                 ]
 
+    if args["--spark-mesos-executor-gpus"]:
+        MEMORY_MULTIPLIER = 20
```
elezar: Would it make sense to also move the construction of these config options to a helper function? e.g. `get_gpu_conf_args`?

scale-tests/batch_test.py
```diff
@@ -123,10 +128,28 @@ def submit_loop(submits_per_min: int, dispatchers: typing.List[typing.Dict], use
...
+        memory = int(args["--spark-mesos-executor-gpus"]) * MEMORY_MULTIPLIER
+        user_conf += ["--conf", "spark.driver.memory={}g".format(str(memory)),
```
elezar: Some questions:
* Why does the memory depend on the number of GPUs?
* Hoe was the `MEMORY_MULTIPLIER` determined?
* Should the driver and executor memory be the same?
* Should we check that `spark.mesos.gpus.max` and `spark.mesos.executor.gpus` are consistent?
* There is no check that these args are specified if required (especially w.r.t `--docker-image` which has no default value).

scale-tests/batch_test.py
```diff
@@ -123,10 +128,28 @@ def submit_loop(submits_per_min: int, dispatchers: typing.List[typing.Dict], use
...
+                      "--conf", "spark.mesos.gpus.max={}".format(args["--spark-mesos-max-gpus"]),
+                      "--conf", "spark.mesos.executor.gpus={}".format(args["--spark-mesos-executor-gpus"]),
+                      "--conf", "spark.mesos.executor.docker.image={}".format(args["--docker-image"]),
+                      "--conf", "spark.mesos.executor.docker.forcePullImage=false"
+                      ]
+        app_url = GPU_IMAGE_RECOGNITION_APP_URL
+    else:
+        app_url = MONTE_CARLO_APP_URL
+
     if args["--spark-mesos-driver-labels"] is not None:
```
elezar: Is:
```python
if args["--spark-mesos-driver-labels"]:
    ...
```
not a better check. This will catch the empty-string case as well.

scale-tests/batch_test.py
```diff
@@ -85,14 +90,14 @@ def submit_job(dispatcher: typing.Dict, duration: int, config: typing.List[str])
     else:
         spark_utils.submit_job(
             service_name=dispatcher_name,
-            app_url=MONTE_CARLO_APP_URL,
+            app_url=app_url,
             app_args=app_args,
             verbose=False,
             args=config,
             driver_role=dispatcher["roles"]["executors"])
```
elezar: I know it wasn't done in this PR, but could we use local variables for `spark_user` and `principal` and set these based on the strict mode. instead of having two calls to the same function.

scale-tests/batch_test.py
elezar: We also don't check whether `--max-num-dispatchers` makes sense for the specified list of dispatchers.

scale-tests/batch_test.py
```diff
@@ -73,15 +78,42 @@ def _is_strict(dispatcher: typing.Dict) -> bool:
...
+            --docker-image
+            """)
+
+    _verify_required_args()
+
+    # Based on testing, 20gb per GPU is needed to run the job successfully.
+    # This is due to memory being divvied up and allocated to each GPU device.
+    memory_multiplier = 20
+    memory = int(args["--spark-mesos-executor-gpus"]) * memory_multiplier
+    return ["--conf", "spark.driver.memory={}g".format(str(memory)),
```
susanxhuynh: @samvantran  Looks like there's a problem: the submit_job() method actually hardcodes the driver memory to 2g.

Also, do you set the driver cpus? In the quota spreadsheet, I noticed it was set to 16 cpus but I don't see it here.

samvantran: Good catch! Fixed

spark-testing/spark_utils.py
```diff
@@ -127,10 +127,10 @@ def submit_job(
 
     conf_args = args.copy()
 
-    conf_args += [
-        '--conf spark.driver.memory=2g',
-        '--conf spark.mesos.role={}'.format(driver_role)
-    ]
+    conf_args += ['--conf spark.mesos.role={}'.format(driver_role)]
+
+    if "spark.driver.memory" not in conf_args:
```
elezar: I don't think that this check makes sense. `conf_args` is a list, and contains, for example:
```python
[ ..., "--conf", "spark.driver.memory=20g", ...]
```
which means that this check will *always* be `False`. We would have to change the check to use something like:
```python
if not list(filter(lambda x: x.startswith("spark.driver.memory=""), conf_args)):
      conf_args += ['--conf spark.driver.memory=2g']
```

spark-testing/spark_utils.py
```diff
@@ -127,10 +127,10 @@ def submit_job(
...
+        conf_args += ['--conf spark.driver.memory=2g']
```
elezar: Considering the check that we are performing above, let's be consistent with how we are adding these args. i.e. let's use:
```python
conf_args += ['--conf', 'spark.driver.memory=2g', ]
```
instead.

scale-tests/batch_test.py
elezar: Is the same amount of memory requires per driver as per executor?

scale-tests/batch_test.py
```diff
@@ -73,15 +78,43 @@ def _is_strict(dispatcher: typing.Dict) -> bool:
...
+            "--conf", "spark.executor.memory={}g".format(str(memory)),
+            "--conf", "spark.cores.max={}".format(args["--spark-cores-max"]),
```
elezar: `spark.cores.max` is already set as part of the "standard" config options.

spark-testing/spark_utils.py
samvantran: dang.. good catch. I missed that one. I'll use your suggestion.

scale-tests/batch_test.py
samvantran: you mean in actual Spark's configuration? It doesn't look like it: https://spark.apache.org/docs/latest/configuration.html#scheduling

unless you meant its set somewhere else?

scale-tests/batch_test.py
samvantran: they don't have to be exactly the same but in my testing in the past, the jobs would fail if the memory was less than 20gb since the docker image itself is very large.

samvantran: @susanxhuynh , you're right. I've never set `spark.driver.cores`. We could probably lower the driver cpu quota a bunch.. but the executors do need 16

scale-tests/batch_test.py
elezar: > What do you mean by consistent? That max is always >= executor.gpus?
Yes. It's not critical though.

PR #348, opened 2018-06-04
Add master script to run entire spark scale tests

samvantran:
STILL WIP

It would be good to have a master script that can run all the spark
scale test commands in one go rather than have 3 different team members
sitting around coordinating the runs.

The idea is to run the script with 
```
./run_spark_scale_test.sh \
  SSH_KEY=/path/to/key \
  CLUSTER_USERNAME=user \
  CLUSTER_PASSWORD=pass \
```
  etc
ee3b04e


scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+then
+  echo "Need to set SECURITY either to 'strict' or 'permissive'"
+  exit 1
+fi
+
+
+## Get a stable shell 
+git clone git@github.com:mesosphere/spark-build.git
+cd spark-build
+docker build -t mesosphere/dcos-commons:${LOGNAME}-spark scale-tests/
```
elezar: Should one not drop into the shell here?

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+./scale-tests/deploy-dispatchers.py \
+  --quota-drivers-cpus $QUOTA_DRIVERS_CPUS \
+  --quota-drivers-mem $QUOTA_DRIVERS_MEM \
+  --quota-executors-cpus $QUOTA_EXECUTORS_CPUS \
+  --quota-executors-mem $QUOTA_EXECUTORS_MEM \
+  $NUM_DISPATCHERS $TEST_NAME ${TEST_NAME}-${NUM_DISPATCHERS}
+
+## Deploy GPU dispatchers
+
+## There is a bug in the script that auto-fails if a dispatcher already exists so
```
elezar: We should rather update `deploy` dispatchers script to fix this issue.

Also, would it not be better to deploy different sets of dispatchers (by changing the prefix, for example).

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+
+## Upload dispatchers file to S3
+aws s3 cp --acl public-read \
+  "${DISPATCHERS_JSON_FILE}" \
+  "s3://${TEST_S3_BUCKET}/${TEST_S3_FOLDER}/${DISPATCHERS_JSON_FILE}"
+
+## Deploy streaming infrastructure
+SERVICE_NAMES_PREFIX="${TEST_NAME}"
+INFRASTRUCTURE_JSON_FILE="${TEST_NAME}-infrastructure.json"
+KAFKA_ZK_CONFIG=${KAFKA_ZK_CONFIG:-scale-tests/configs/kafka-zookeeper-options.json}
```
elezar: Does it make sense to specify defaults for other envvars too?

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+KAFKA_CONFIG=${KAFKA_CONFIG:-scale-tests/configs/kafka-options.json}
+CASSANDRA_CONFIG=s${CASSANDRA_CONFIG:-cale-tests/configs/cassandra-options.json}
```
elezar: typo: `-cale-tests`.

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+  --kafka-zookeeper-config $KAFKA_ZK_CONFIG \
+  --kafka-cluster-count $NUM_KAFKA_CLUSTERS \
+  --kafka-config $KAFKA_CONFIG \
+  --cassandra-cluster-count $NUM_CASSANDRA_CLUSTERS \
+  --cassandra-config $CASSANDRA_CONFIG
+
+
+## Upload infrastructure file to S3
+aws s3 cp --acl public-read \
+  "${INFRASTRUVTURE_JSON_FILE}" \
```
elezar: typo.

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+  "s3://${TEST_S3_BUCKET}/${TEST_S3_FOLDER}/${INFRASTRUCTURE_JSON_FILE}"
+
+## Prepare dispatcher list files
+NUM_INFINITE_DISPATCHERS="$(expr "${NUM_DISPATCHERS}" - "${NUM_FINITE_DISPATCHERS}")"
+DISPATCHERS_FINITE_JSON_FILE="${TEST_NAME}-finite-dispatchers.json"
```
elezar: Rather launch two different sets of dispatchers than having to manipulate the list of dispatchers.

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+
+./scale-tests/kafka_cassandra_streaming_test.py \
+  $DISPATCHERS_INFINITE_JSON_FILE \
+  $INFRASTRUCTURE_JSON_FILE \
+  $SUBMISSIONS_FINITE_JSON_FILE \
+  --jar $KAFKA_JAR \
+  --num-producers-per-kafka $NUM_PRODUCERS_PER_INFINITE_KAFKA \
+  --num-consumers-per-producer $NUM_CONSUMERS_PER_INFINITE_PRODUCER \
+  --producer-number-of-words $INFINITE_PRODUCER_NUMBER_OF_WORDS \
+  --producer-words-per-second $NUM_PRODUCER_WORDS_PER_SECOND \
```
elezar: We could drop this argument. It does not make it clearer to read the script, and `kafka_cassandra_streaming_test.py` has a default value of 0.

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+  --producer-spark-cores-max $PRODUCER_SPARK_CORES_MAX \
+  --producer-spark-executor-cores $PRODUCER_SPARK_EXECUTOR_CORES \
+  --consumer-batch-size-seconds $CONSUMER_BATCH_SIZE_SECONDS \
+  --consumer-spark-cores-max $CONSUMER_SPARK_CORES_MAX \
+  --consumer-spark-executor-cores $CONSUMER_SPARK_EXECUTOR_CORES
+
+## Deploy GPU jobs
+git checkout st-gpu-strict
```
elezar: Be careful changing this mid-script. Rather ensure that the branch we are running is consistent.

Ideally, for scripts such as this, we should *only* run from `master` (or some "stable" branch).

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+nohup ./scale-tests/batch_test.py $DISPATCHERS_JSON_FILE \
```
elezar: This is the wrong dispatchers file? (or was there another reason for creating a different set of GPU dispatchers?)

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+    --spark-mesos-max-gpus 4 \
+    --no-supervise &
+
+git checkout master
+
+## Deploy batch jobs via marathon app
+./scale-tests/batch_test.py ${DISPATCHERS_JSON_FILE} --submits-per-min 4
+
+# The following is commented out due to a bug in the marathon app
+# Also, I'm not sure how to upload files to s3 in here
```
elezar: See the `aws s3 cp` commands above.

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+
+git checkout master
+
+## Deploy batch jobs via marathon app
+./scale-tests/batch_test.py ${DISPATCHERS_JSON_FILE} --submits-per-min 4
+
+# The following is commented out due to a bug in the marathon app
+# Also, I'm not sure how to upload files to s3 in here 
+#
+#cd scale-tests/
```
elezar: In #347 this script has been changed to a Python script. We should also run from the same folder if we can help it.

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+#   6 13000 "${DISPATCHER_JSON_FILE} --submits-per-min 4"
+# cd ..
+
+###### EOF
+
+## Run script inside stable container
+docker exec --net=host \
+  -v $(pwd):/spark-build \
+  -v $SSH_KEY:/ssh/key:ro \
+  -v ~/.aws./credentials:~/.aws/credentials \
```
elezar: When mounting file paths to docker, absolute paths must be used. See the way the credential file is mounted into `test.sh`, for example. It may also be a good idea to mount this `:ro` (read-only).

scale-tests/run_spark_scale_test.sh
```diff
@@ -0,0 +1,226 @@
...
+## Add Spark stub
+dcos package repo add --index=0 spark-aws https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts.s3.amazonaws.com/permanent/spark/assets/scale-testing/stub-universe-spark.json
+
+## Deploy dispatchers
+./scale-tests/deploy-dispatchers.py \
+  --quota-drivers-cpus $QUOTA_DRIVERS_CPUS \
+  --quota-drivers-mem $QUOTA_DRIVERS_MEM \
+  --quota-executors-cpus $QUOTA_EXECUTORS_CPUS \
+  --quota-executors-mem $QUOTA_EXECUTORS_MEM \
+  $NUM_DISPATCHERS $TEST_NAME ${TEST_NAME}-${NUM_DISPATCHERS}
```
susanxhuynh: One thing to be careful of is this script currently does not wait for the dispatchers to finish deploying (see "wait_for_deployment=False" in deploy-dispatchers.py). You could change this option while installing, or check the dispatchers are ready right before the job submission. With the former, you might want to check if how long this takes since it would be deploying the dispatchers serially.

mpereira:
We may want to upstream the sdk_install changes. Maybe make the auto-fail optional.

elezar:
@mpereira this seems to contain the changes from #345 which still has some open comments.

testing/sdk_install.py
```diff
@@ -90,10 +90,6 @@ def install(
         insert_strict_options=True):
     start = time.time()
 
-    # If the package is already installed at this point, fail immediately.
-    if sdk_marathon.app_exists(service_name):
```
elezar: Is this really what we want? We need to be careful making changes to `testing/` that are not ported to `dcos-commons` as these will be overwritten the next time we update the tooling.

mpereira:
@elezar ah, good spot.

testing/sdk_marathon.py
```diff
@@ -132,7 +132,7 @@ def install_app(app_definition: dict) -> (bool, str):
     app_name = app_definition["id"]
 
     with tempfile.TemporaryDirectory() as d:
-        app_def_file = "{}.json".format(app_name)
+        app_def_file = "{}.json".format(app_name.replace('/', '__'))
```
elezar: See note above on making these changes outside of `dcos-commons`. An alternative would be to create the path to the file.

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
...
+read -p "Run GPU batch jobs? [y/N]: " ANSWER
+case "${ANSWER}" in
+  [Yy]* ) SHOULD_RUN_GPU_BATCH_JOBS=true;;
+  * ) ;;
+esac
+
+if docker top "${CONTAINER_NAME}" > /dev/null 2>&1; then
+  log 'Container already running'
+else
+  git clone git@github.com:mesosphere/spark-build.git "${TEST_DIRECTORY}" | tee -a "${LOG_FILE}"
```
porridge: Why do we need this directory? Why not just use `$(pwd)`?

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
...
+esac
+
+echo
+read -p "Run GPU batch jobs? [y/N]: " ANSWER
+case "${ANSWER}" in
+  [Yy]* ) SHOULD_RUN_GPU_BATCH_JOBS=true;;
+  * ) ;;
+esac
+
+if docker top "${CONTAINER_NAME}" > /dev/null 2>&1; then
```
porridge: This is not idempotent, if any container initialization step fails after `docker run`, then it will not be re-run.

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
```
porridge: Should we also `set -x` to make it possible/easy to re-run failed steps manually?

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
...
+function log {
+  local -r message="${*:-}"
+  echo "$(date "+%Y-%m-%d %H:%M:%S") | ${message}" 2>&1 | tee -a "${LOG_FILE}"
+}
+
+function container_exec () {
+  local -r command="${*}"
+  local -r full_command="
+    . ${CONTAINER_SSH_AGENT_EXPORTS}
+    ssh-add -k ${CONTAINER_SSH_KEY} > /dev/null 2>&1
```
porridge: Why re-add the key with every command, rather than just at the beginning?

testing/sdk_install.py
```diff
@@ -90,10 +90,6 @@ def install(
         insert_strict_options=True):
     start = time.time()
 
-    # If the package is already installed at this point, fail immediately.
```
porridge: I'm not sure we should be changing the copied `testing/` subdirectory in the same PR as changing local code.

testing/sdk_install.py
samvantran: Why do we have this to begin with? Was it for CI/soak?

Without this, it is difficult to re-run the scale tests if there is a failure somewhere in the chain. But as @mpereira suggested, it might be better to include a flag so that auto-fail is optional.

mpereira: I think in the absence of a thorough verification that a similarly-named service is being installed has the same configuration, failing immediately might avoid some possible confusion.

scale-tests/run.sh
elezar: Because this is running in a container.

scale-tests/run.sh
elezar: Would using `docker ps` not be a better solution here?

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
...
+  log 'Running infinite jobs. Consumers do not write to Cassandra'
+  container_exec \
+    ./scale-tests/kafka_cassandra_streaming_test.py \
+      "${DISPATCHERS_JSON_OUTPUT_FILE}" \
+      "${INFRASTRUCTURE_OUTPUT_FILE}" \
+      "${INFINITE_SUBMISSIONS_OUTPUT_FILE}" \
+      --jar "${TEST_ASSEMBLY_JAR_URL}" \
+      --num-producers-per-kafka "${INFINITE_NUM_PRODUCERS_PER_KAFKA}" \
+      --num-consumers-per-producer "${INFINITE_NUM_CONSUMERS_PER_PRODUCER}" \
+      --producer-number-of-words "${INFINITE_PRODUCER_NUMBER_OF_WORDS}" \
```
elezar: `INFINITE_PRODUCER_NUMBER_OF_WORDS` should *always* be 0.

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
...
+      --producer-spark-executor-cores "${INFINITE_PRODUCER_SPARK_EXECUTOR_CORES}" \
+      --consumer-batch-size-seconds "${INFINITE_CONSUMER_BATCH_SIZE_SECONDS}" \
+      --consumer-spark-cores-max "${INFINITE_CONSUMER_SPARK_CORES_MAX}" \
+      --consumer-spark-executor-cores "${INFINITE_CONSUMER_SPARK_EXECUTOR_CORES}"
+
+  log 'Uploading infinite jobs submissions file'
+  container_exec \
+    aws s3 cp --acl public-read \
+      "${INFINITE_SUBMISSIONS_OUTPUT_FILE}" \
+      "s3://${TEST_S3_BUCKET}/${TEST_S3_FOLDER}/"
```
elezar: `"s3://${TEST_S3_BUCKET}/${TEST_S3_FOLDER}/"` is used repeatedly. Should we have a single envvar for this?

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
...
+      --dcos-password "${DCOS_PASSWORD}" \
+      --security "${SECURITY}" \
+      --input-file-uri "${DISPATCHERS_JSON_OUTPUT_FILE_URL}" \
+      --script-cpus "${BATCH_SCRIPT_CPUS}" \
+      --script-mem "${BATCH_SCRIPT_MEM}" \
+      --script-args "\"\
+        ${DISPATCHERS_JSON_OUTPUT_FILE} \
+        --submits-per-min ${BATCH_SUBMITS_PER_MIN} \
+      \"" \
+      --spark-build-branch "${BATCH_SPARK_BUILD_BRANCH}"
```
elezar: Nit: Let's keep the `--script-args` as the last argument.

scale-tests/run.sh
```diff
@@ -0,0 +1,434 @@
...
+        ${GPU_DISPATCHERS_JSON_OUTPUT_FILE} \
+        --submits-per-min ${GPU_SUBMITS_PER_MIN} \
+        --docker-image ${GPU_DOCKER_IMAGE} \
+        --max-num-dispatchers ${GPU_MAX_NUM_DISPATCHERS} \
+        --spark-cores-max ${GPU_SPARK_CORES_MAX} \
+        --spark-mesos-executor-gpus ${GPU_SPARK_MESOS_EXECUTOR_GPUS} \
+        --spark-mesos-max-gpus ${GPU_SPARK_MESOS_MAX_GPUS} \
+        --no-supervise \
+      \"" \
+      --spark-build-branch "${GPU_SPARK_BUILD_BRANCH}"
```
elezar: Nit: Same comment regarding `--script-args`.

scale-tests/run.sh
porridge: Why does it matter? `ssh-agent` keeps running in that container when it's started. It will keep the key loaded for as long as it's running, there is no point re-adding that key a million times.

mpereira: @porridge you're right. The sourcing of the ssh-agent environment variables (line 80, `. ${CONTAINER_SSH_AGENT_EXPORTS}`) is enough to wire the shell instance with the ssh-agent session.

scale-tests/run.sh
mpereira: I found `set -x` to be too verbose given the script logging. `container_exec` logs all commands (with the timestamp) for example.

scale-tests/run.sh
mpereira: @porridge https://github.com/mesosphere/spark-build/pull/348/commits/de5fb303a534d11282f806b3de5f2aaaf56d0166

scale-tests/run.sh
mpereira: https://github.com/mesosphere/spark-build/pull/348/commits/330ade634f6f9bbbb195c6249f10299a2bdf5a0e

scale-tests/run.sh
mpereira: https://github.com/mesosphere/spark-build/pull/348/commits/aa0310b9457aa0a421e93eb1405042c3d981353f

scale-tests/run.sh
mpereira: https://github.com/mesosphere/spark-build/pull/348/commits/aa0310b9457aa0a421e93eb1405042c3d981353f

scale-tests/run.sh
mpereira: https://github.com/mesosphere/spark-build/pull/348/commits/5675d037074f934f4936177fef26f2bedf919150

scale-tests/run.sh
mpereira: @elezar the nice thing about `docker top` is that it returns 1 if the container isn't running, so it's terser to use in conditionals (e.g. no need to parse the output of `docker ps`). Do you think we should still use `docker ps`?

mpereira: Maybe this is better? https://github.com/mesosphere/spark-build/pull/348/commits/98754682c941196d0c1651a7643588e7ae1122a7

scale-tests/run.sh
mpereira: We want to have reproducible test runs, I'll push a thing in a bit: https://github.com/mesosphere/spark-build/pull/348#issuecomment-396608807

mpereira:
Thanks for pushing https://github.com/mesosphere/spark-build/pull/348/commits/66f144411398384fa158b0db092a8a614de4fe30 @susanxhuynh, I had the change locally but forgot to push it.

mpereira:
@susanxhuynh @samvantran @elezar this fixes the Marathon installs getting stuck https://github.com/mesosphere/spark-build/pull/348/commits/d555df06fdcabedfd940952a5a23dfff072b4dda

I'll push the `/testing` changes made here to dcos-commons too.

elezar:
@mpereira with regards to:
> @elezar yes I had that in mind but didn't implement it. There's a bit of a "chicken and egg" condition with it which is run.sh will need to be executed to check for the for the repo commit hash which should be used for run.sh. Maybe it makes sense having the commit hash in the config file and making run.sh fail if HEAD differs (and/or if the repo is dirty).
>
> I'll think about it and push in a bit.

Yes, I'm not to sure how to deal with this. We could leave it as is for the time being, or at the very least log the information?

testing/sdk_install.py
elezar: cc @mpereira. Could we revert this change? This would cause a lot of silent failures in our `dcos-commons` CI?

mpereira: https://github.com/mesosphere/spark-build/pull/348/commits/23c19ef3f3a0b01482cd74396d616d5bdd990239

mpereira:
dcos-commons testing port PR: https://github.com/mesosphere/dcos-commons/pull/2528

mpereira:
@elezar I have a [WIP branch](https://github.com/mesosphere/spark-build/compare/st-master-script...mpereira/master-script?expand=1) for introducing git refs as a script argument, where I also want to move some of the script arguments to configuration variables (as @porridge suggested). Are you ok with merging this PR as is and getting the other one ready for the next test?

PR #349, opened 2018-06-06
[DCOS-34447] Export kafka consumer metrics from the client app.

porridge:
Among others, this makes kafka_consumer_fetch_manager_metrics_records_lag_max
available, which looks like a reasonable candidate to resolve DCOS-34447.


porridge:
FTR the tip of this branch is built as http://infinity-artifacts.s3.amazonaws.com/scale-tests/dcos-spark-scala-tests-assembly-mowsiany-20180606-0071423.jar in case someone wants to try it out

PR #350, opened 2018-06-11
[DCOS-37663] fixed spark security setup error for strict mode

susanxhuynh:
We missed a few things after the recent dcos-commons tools upgrade that caused tests to fail in strict mode.

- Change in sdk_security.grant() interface.
- Switch from app_id to service_name => removing the leading slash.


susanxhuynh:
I ran both the permissive and strict mode CI jobs against this PR. Both passed.

spark-testing/spark_utils.py
```diff
@@ -26,7 +26,7 @@
 SPARK_SERVICE_ACCOUNT = os.getenv("SPARK_SERVICE_ACCOUNT", "spark-service-acct")
 SPARK_SERVICE_ACCOUNT_SECRET = os.getenv("SPARK_SERVICE_ACCOUNT_SECRET", "spark-service-acct-secret")
 SPARK_SERVICE_NAME = os.getenv("SPARK_SERVICE_NAME", "spark")
-FOLDERED_SPARK_SERVICE_NAME = "/path/to/" + SPARK_SERVICE_NAME
+FOLDERED_SPARK_SERVICE_NAME = "path/to/" + SPARK_SERVICE_NAME
```
elezar: In `dcos-commons` we explicitly have a leading `/`. There may be another reason why the tooling is causing this to fail.

spark-testing/spark_utils.py
```diff
@@ -217,20 +217,17 @@ def spark_security_session():
     service_account = SPARK_SERVICE_ACCOUNT
     secret = SPARK_SERVICE_ACCOUNT_SECRET
 
-    def grant_driver_permission(service_account_name, app_id):
-        dcosurl, headers = sdk_security.get_dcos_credentials()
+    def grant_driver_permission(service_account_name, service_name):
+        app_id = "/{}".format(service_name)
```
elezar: Instead of removing the `/` from `FOLDERED_SERVICE_NAME`, let's use:
```python
app_id = "/{}".format(service_name.lstrip("/"))
```
to ensure that `app_id` only starts with a single slash.

susanxhuynh: 👍

susanxhuynh:
@elezar I've made the change that you suggested.

PR #352, opened 2018-06-12
[DCOS-34447] Make kafka metrics work on executors.

porridge:
- use a patched version of reporter which defaults to environment
  for connection settings - built from
  https://github.com/porridge/kafka-statsd-metrics2/tree/from-env-with-version-bump
- drop the host/port from kafka properties to let reporter use the
  settings in environment.


tests/jobs/scala/build.sbt
```diff
@@ -12,7 +12,7 @@ lazy val root = (project in file("."))
       "org.apache.kafka" % "kafka_2.11" % "0.10.0.1",
       "com.github.scopt" %% "scopt" % "3.7.0",
       "com.datastax.spark" %% "spark-cassandra-connector" % "2.0.7",
-      "com.airbnb" % "kafka-statsd-metrics2" % "0.5.3" from "http://downloads.mesosphere.com/kafka/assets/kafka-statsd-metrics2-0.5.3.jar",
+      "com.airbnb" % "kafka-statsd-metrics2" % "0.5.2+issue28" from "https://infinity-artifacts.s3.amazonaws.com/scale-tests/kafka-statsd-metrics2-0.5.2+issue28-20180612-8c1e3c4f3fa83.jar",
```
elezar: Not that it's critical, but we could consider moving this `jar` to a permanent `assets` folder.

porridge:
Thanks.

I'd like to wait with moving until I hear back from airbnb guys about whether my "fix" is OK.

This does not change behavior for anything but the executors, so it would only make sense to update other packages for consistency sake, but see the previous point.

PR #353, opened 2018-06-12
[DCOS-34447] Use assembly jar with working executor kafka client metrics.

porridge:


porridge:
#354 is against a different branch

PR #354, opened 2018-06-12
[DCOS-34447] Use assembly jar with working executor kafka client metrics.

porridge:


porridge:
cc @samvantran @susanxhuynh - please make sure to use this jar in the next round of scale tests

mpereira:
We should use the new JAR for the next tests but I think we should keep the existing test configs unchanged as history.

susanxhuynh:
👍 Ok. Let's create a new set of configs for the next test like Murilo suggested.

porridge:
Done, copies made, changes to old configs reverted.
PTAL

PR #355, opened 2018-06-12
DCOS-38168: Deploy dispatchers concurrently.

mpereira:


susanxhuynh:
Just curious, what is the performance like? Did you time it with a certain number of dispatchers? If you have time, could you test with 50 dispatchers (or maybe up to 20 on a small cluster)?

susanxhuynh:
(I'm wondering if that many concurrent threads will run successfully on a laptop.)

scale-tests/deploy-dispatchers.py
```diff
@@ -63,6 +63,8 @@
 
 log = logging.getLogger(__name__)
 
+MAX_THREADPOOL_WORKERS = 50
```
elezar: Should this be configurable, either through an envvar or a command-line option?

scale-tests/deploy-dispatchers.py
```diff
@@ -236,20 +228,24 @@ def deploy_dispatchers(
...
-                                                linux_user,
-                                                lambda x: 0,
-                                                None,
-                                                options,
-                                                quota_options)
-            dispatchers.append(dispatcher_settings)
-            outfile.write("{},{},{}\n".format(dispatcher_settings["service"]["name"],
-                                              dispatcher_settings["roles"]["drivers"],
-                                              dispatcher_settings["roles"]["executors"]))
+    def deploy_dispatcher(i):
```
elezar: We could add type annotations, and I'm on the fence about whether we need to use a different parameter name than `i`. (`index` perhaps?)

scale-tests/deploy-dispatchers.py
```diff
@@ -236,20 +228,24 @@ def deploy_dispatchers(
...
+                               lambda x: 0,
+                               None,
+                               options,
+                               quota_options)
+
+    with ThreadPoolExecutor(max_workers=MAX_THREADPOOL_WORKERS) as executor:
+        dispatchers = list(executor.map(deploy_dispatcher, range(num_dispatchers)))
+
+    with open(output_file, 'w') as outfile:
+        for dispatcher in dispatchers:
```
elezar: Should we sort the dispatchers? Their order will now be non-deterministic, correct?

scale-tests/deploy-dispatchers.py
elezar: I also don't know if it's required to be set here. We could just use a local variable were we instantiate the `ThreadPoolExecutors`.

scale-tests/deploy-dispatchers.py
mpereira: The generator returned by `executor.map()` yields dispatchers in the same order of the collection mapped over.

scale-tests/deploy-dispatchers.py
mpereira: > Should this be configurable, either through an envvar or a command-line option?

I thought about doing that but decided otherwise. Do you think we should?

mpereira: > I also don't know if it's required to be set here. We could just use a local variable were we instantiate the ThreadPoolExecutors.

That's just my personal preference for setting "magic values" as really visible constants instead of inside functions. If you feel strongly about it I can change it.

mpereira:
> Just curious, what is the performance like? Did you time it with a certain number of dispatchers? If you have time, could you test with 50 dispatchers (or maybe up to 20 on a small cluster)?

@susanxhuynh I added some timing calls to the script and deploying 20 dispatchers on kafka110s took 5 minutes:
```
2018-06-13 11:27:15 | Installed non-GPU dispatchers in 292 seconds
```

mpereira:
Timing calls were added here https://github.com/mesosphere/spark-build/pull/348/commits/30d1ee10cc55ff78dd9a3c6651130769e98fdb35

mpereira:
> (I'm wondering if that many concurrent threads will run successfully on a laptop.)

50 threads doing mostly IO work should run fine.

susanxhuynh:
OK, thanks for checking the performance.

scale-tests/deploy-dispatchers.py
elezar: Let's make it configurable if and when we require it. I'm ok with it as it is.

samvantran:
Nice work !

PR #357, opened 2018-06-13
[DCOS-38141] Add SDK dashboard for scaletesting

elezar:
This PR adds a script to generate a Grafana dashboard for SDK-based services and adds an example dashboard.

Note that the dashboard "automatically" detects if SDK services are running and allows for the framework to be selected.

Note that `ROOT`/`MOUNT` volumes are not currently supported by this dashboard.

screen shot 2018-06-13 at 17 03 01


scale-tests/sdk.dashboard.py
```diff
@@ -0,0 +1,266 @@
...
+    Construct a reduction string for a metric.
+    """
+    base = "{}({})".format(op, metric_definition)
+    if by:
+        return "{} by ({})".format(base, by)
+
+    return base
+
+
+def metric(name: str, selection: str) -> str:
```
porridge: Consider making selection a dict for a more pythonic invocation syntax.

scale-tests/sdk.dashboard.py
```diff
@@ -0,0 +1,266 @@
...
+pip install grafanalib
+
+generate-dashboard -o sdk.json sdk.dashboard.py
+```
+
+This creates an `sdk.json` file that contains the Grafana dashboard definition for monitoring SDK-based frameworks
+installed on a Cluster.
+"""
+
+from grafanalib.core import *
```
porridge: Please enumerate the pieces you need. `*` is considered bad style because it's a maintenance headache.

scale-tests/sdk.dashboard.py
```diff
@@ -0,0 +1,266 @@
...
+                span=2,
+                targets=[Target(expr=sum(service_metric(m, "rate", resolution)), legendFormat=m) for m in task_metrics],
+                steppedLine=True,
+                yAxes=YAxes(left=YAxis(format="short", decimals=0))
+            ),
+        ]
+    )
+
+
+# Generate the Grafana dashboard.
```
porridge: This comment is not very useful. It's clear from the subsequent line that you're creating a dashboard...

scale-tests/sdk.dashboard.py
```diff
@@ -0,0 +1,266 @@
+"""
+Generate a generic SDK dashboard.
+
+```bash
+pip install grafanalib
+
+generate-dashboard -o sdk.json sdk.dashboard.py
```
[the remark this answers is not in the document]
elezar: From: https://github.com/weaveworks/grafanalib#generating-dashboards
> If you save the above as frontend.dashboard.py (the suffix must be .dashboard.py), you can then generate the JSON dashboard with:

porridge:
Ah, that's interesting. Mind linking to that page there?

elezar:
Thanks @porridge and @samvantran for the review. I have addressed the issues you raise.

@samvantran with regards to:
> I ran generate-dashboard -o sdk.json sdk.dashboard.py and uploaded the generated json to Grafana and this is what I saw:

You need to select a `framework` from the drop-down list at the top of the page. It could be that the metrics are not present to populate the list, but this has worked every time that I have tested it.

PR #358, opened 2018-06-15
Some fixes from 0614 test

susanxhuynh:
- made the service account private and public keys unique so they wouldn't collide during parallel execution
- don't install Spark CLI in threads because they will all install into the same file name and clash. also reduced the number of threads in pool, making it easier to see the logs, deployment of 50 dispatchers took 15 min.


testing/sdk_security.py
```diff
@@ -159,17 +159,21 @@ def create_service_account(service_account_name: str, service_account_secret: st
     log.info('Remove any existing service account and/or secret')
     delete_service_account(service_account_name, service_account_secret)
 
+    private_key_file = "{}-private-key.pem".format(service_account_name)
```
porridge: Please instead use a separate `with tempfile.TemporaryDirectory(...)` for each pair, this will also take care of cleaning them up.

elezar: Thanks @porridge. Yes, in hindsight, that should have been done in the `sdk_security` improvements I pushed through recently.

mpereira:
> also reduced the number of threads in pool, making it easier to see the logs

@susanxhuynh do you mean the shell output or the [log file](https://github.com/mesosphere/spark-build/pull/348/files#diff-5f81cedb7de857f3a6139a9d6ed6d81bR67) the script pipes the shell output to?

If you mean the former, using the log file should be better for following output.

testing/sdk_security.py
elezar: Also, please create a ticket to add these changes to `sdk_security` in `dcos-commons`.

susanxhuynh: OK, will do.

susanxhuynh: @porridge I've added `with tempfile.TemporaryDirectory()`.
@elezar Ticket to port this to dcos-commons: https://jira.mesosphere.com/browse/DCOS-38316

susanxhuynh:
@mpereira Good suggestion. I've reverted the max threads back to 50.

PR #361, opened 2018-06-18
[DCOS-28729, DCOS-34448] Add metric for ZK node health.

porridge:


porridge:
I don't think I understand your question.
FTR, this is how the graph looks like on my TV for the duration of the test:
health

elezar:
I mean that we are not able to recognise an unhealthy cluster by just glancing at the graph, because if a ZK ensemble has 3 or fewer servers, it will not plot anything.

porridge:
Pushed another commit adding a graph for healthy kafka and ZK scheduler count.

health2

porridge:
If an ensemble goes to 3, then the line for 4 or 5 will drop from the target of 25 to something lower.
Recognizing which one is in trouble is out of scope for this dashboard in general (at least for now).

porridge:
Pushed a fix for the ensemble graph. PTAL

elezar:
> The reason for having ==5 and >=4 is so that it's easy to see whether we're meeting the SLO or not in both cases - the thresholds added to the graph mean that all you need to watch is that the line is in the green background area.

Yes, after re-reading the tooltip for the graph I think I'm starting to agree with this more. Good work, by the way.

PR #373, opened 2018-07-04
[DCOS-38388] A script to define a metric which declares current scale test phase.

porridge:
This can be used by prometheus expressions to select alert/graph
thresholds.


porridge:
I do plan to write some over-arching documentation once I have all the pieces in the puzzle, in particular:
 - without an instance of https://github.com/mesosphere/prometheus/pull/19 this script is nearly useless
 - I also need to write some actual alert expressions using this metric to confirm that my idea is sound.

As for language choice, it's mostly that I happened to have a tested way to run curl with a socks proxy in bash, and I'd have to research a bit how to do it in Python. Also, There is no clear line but I tend to switch to Python once there is some more complex processing involved. Other than argument parsing and diagnostics, this script is really 9 lines of relatively simple logic, so I thought it makes sense to stick with bash. I definitely could port it to Python if you feel strongly about this.

PR #374, opened 2018-07-06
[DCOS-38385] Use graphs instead of metrics.

porridge:


porridge:
Screenshot:
graphs-demo

scale-tests/sdk.dashboard.py
```diff
@@ -132,29 +97,52 @@ def resource_row(pod_type: str = None) -> G.Row:
...
+                title="Memory",
+                expressions=[
+                    dict(expr=reduction('sum', metric(m, selection), selection),
+                         legendFormat=title)
+                    for m, title in [('mem_total', 'Used'),
+                                     ('mem_limit', 'Available')]
+                ],
+                span=3,
+                steppedLine=True,
+                yAxes=G.YAxes(left=G.YAxis(format="short", decimals=0)),
```
elezar: Should we set the units for memory and disk to bytes?

porridge: OK, done.

porridge:
I do not have an explanation for the jagged graph. I suspect this is a problem in the metrics pipeline where labels vanish and reappear on timeseries. I didn't investigate though.

PR #377, opened 2018-07-10
[DCOS-39078] Move alerts into this repo and parametrize the kafka cluster count.

porridge:


porridge:
Merging so that we can provide the file to SRE. Please review when you have a moment @elezar .

PR #378, opened 2018-07-11
Fix Spark summary graph jobsByName

samvantran:
Since Kafka jobs now have unique Ids, this graph needed adjustment otherwise you'd have:
- dozens of Kafka graphs and only 
- 1 graph for all monte-carlo jobs and
- 1 graph for image-recognition jobs

Changes this
screen shot 2018-07-10 at 5 22 47 pm

to this
screen shot 2018-07-10 at 5 26 51 pm

### DIFF
--- a/monitoring/dashboards/SparkJobsSummary.json
+++ b/monitoring/dashboards/SparkJobsSummary.json
@@ -1,34 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "prometheus",
-      "description": "",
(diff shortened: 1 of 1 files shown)


monitoring/dashboards/SparkJobsSummary.json
```diff
@@ -1,34 +1,4 @@
 {
-  "__inputs": [
```
elezar: Actually, why is the DataSource removed?

elezar:
We need to re-open this PR.

PR #380, opened 2018-07-11
Fix Spark summary graph Jobs By Name

elezar:
This reopens #378.

Since Kafka jobs now have unique Ids, this graph needed adjustment otherwise you'd have:
- 100 Kafka graphs and only 
- 1 graph for all monte-carlo jobs and
- 1 image-recognition jobs.


monitoring/dashboards/SparkJobsSummary.json
```diff
@@ -1,34 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
```
elezar: The data source should not be removed.

elezar:
Thanks @samvantran. I see the graph in the top left is empty, but if this is not an issue, then please merge. Note that I cannot approve because this is now seen as my PR.

PR #381, opened 2018-07-11
Fix janitor zk node path.

mpereira:
The ZK node path for spark dispatchers was incorrect for foldered service names.

This is an example of an incorrect path for a foldered spark dispatcher:
```
spark_mesos_dispatcher2018-07-11-murilo-0/spark-00
```

When it should look like this:

```
spark_mesos_dispatcher2018-07-11-murilo-0__spark-00
```


samvantran:
LGTM

PR #382, opened 2018-07-12
[DCOS-38374] Add prom spark alerts 

samvantran:
Resolves: [DCOS-38374](https://jira.mesosphere.com/browse/DCOS-38374)

These are the alerts I created for the last large scale test on 7-11-18.


monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
+groups:
+- name: Spark_Alerts
+  rules:
+    - alert: Spark_Dispatcher_Count
+      expr: count(spark_mesos_cluster_driver_launched) >= 50
+      for: 1m
+      labels:
+        severity: major
```
porridge: why not informational?

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      annotations:
+        description: 'Spark dispatcher count unexpectedly decreasing'
+
+    - alert: Spark_Job_Count
+      expr: count(spark_mesos_cluster_executor_count) >= 1000
+      for: 1m
+      labels:
+        severity: major
```
porridge: Ditto.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      annotations:
+        description: 'Spark job count unexpectedly decreasing'
+
+    - alert: Spark_CPU_Job
+      expr: count(spark_mesos_cluster_executor_count{task_name=~".*monte.*"}) >= 750
+      for: 1m
+      labels:
+        severity: major
```
porridge: Ditto.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      annotations:
+        description: 'Spark has reached desired number of CPU jobs'
+
+    - alert: Spark_GPU_Jobs
+      expr: count(spark_mesos_cluster_executor_count{task_name=~".*image.*"}) >= 10
+      for: 1m
+      labels:
+        severity: major
```
porridge: Ditto.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      annotations:
+        description: 'Spark has reached desired number of GPU jobs'
+
+    - alert: Spark_Streaming_Jobs
+      expr: count(spark_mesos_cluster_executor_count{task_name=~".*Kafka.*"}) >= 250
+      for: 1m
+      labels:
+        severity: major
```
porridge: Ditto.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      annotations:
+        description: 'Spark has reached desired number of Streaming jobs'
+    
+    - alert: Spark_Average_Queue_Depth
+      expr: avg(spark_mesos_cluster_driver_waiting) >= 50
+      for: 3m
+      labels:
+        severity: major
+      annotations:
+        description: 'Spark getting backed up'
```
porridge: I wouldn't know what this means.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+    - alert: Spark_Streaming_Jobs
+      expr: count(spark_mesos_cluster_executor_count{task_name=~".*Kafka.*"}) >= 250
+      for: 1m
+      labels:
+        severity: major
+      annotations:
+        description: 'Spark has reached desired number of Streaming jobs'
+    
+    - alert: Spark_Average_Queue_Depth
+      expr: avg(spark_mesos_cluster_driver_waiting) >= 50
```
porridge: What is the meaning of 50 here?

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      labels:
+        severity: major
+      annotations:
+        description: 'Spark job launch time has high variance'
+
+    - alert: Spark_Job_Success_Rate
+      expr: (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) / (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_lost) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_failed))) >= 99.0
+      for: 1m
+      labels:
+        severity: major
```
porridge: Also why not informational?

monitoring/alerts/spark-rules.yml
samvantran: no reason other than I didn't know that was a better alternative. I'll change all the 'good' alerts to informational.

monitoring/alerts/spark-rules.yml
samvantran: The number of jobs queued for each dispatcher. Over 50 means a lot of jobs are building up and not being cleared out fast enough.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      annotations:
+        description: 'Spark has reached desired number of Streaming jobs'
+    
+    - alert: Spark_Average_Queue_Depth
+      expr: avg(spark_mesos_cluster_driver_waiting) >= 50
+      for: 3m
+      labels:
+        severity: major
+      annotations:
+        description: 'Average number of spark jobs queued for each dispatcher growing rapidly, indicating slower launch rates for new jobs or current jobs taking longer to complete'
```
porridge: The message suggests that you're measuring the rate of queue length growth, while in fact the expression just looks at queue length in absolute terms. Moreover, the average is not the best indicator. Please change it to `max()` or 99th percentile, and change the message to say something like "[Longest/99th %ile] spark job queue is over X times longer than expected, indicating..."

samvantran: Actually, after re-reviewing the ticket it looks like product wanted:
> MUST have a queue depth of between 50 and 200 drivers for 75% of the overall test duration per Dispatcher

So I'm changing the alert to an individual value over 200 queued jobs to indicate that it is beyond product limits.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+    - alert: Spark_Launch_Time_Stdev
+      expr: (stddev(spark_mesos_cluster_executor_start_to_all_launched_mean) / 1000) > 20.0
+      for: 3m
+      labels:
+        severity: major
+      annotations:
+        description: 'Spark job launch time has unusually high variance'
+
+    - alert: Spark_Job_Success_Rate
+      expr: (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) / (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_lost) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_failed))) >= 99.0
```
elezar: Is this automatically a percentage? Also, should we not rather (or at least *also*) alert on failure to reach this rate?

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      annotations:
+        description: 'Spark jobs taking longer to launch than expected'
+
+    - alert: Spark_Launch_Time_Stdev
+      expr: (stddev(spark_mesos_cluster_executor_start_to_all_launched_mean) / 1000) > 20.0
+      for: 3m
+      labels:
+        severity: major
+      annotations:
+        description: 'Spark job launch time has unusually high variance'
```
elezar: I may be being a little pedantic, but this description is not really clear. Firstly it is not clear what needs to be done about this, and also variance and standard deviation are not the same thing.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+    - alert: Spark_Average_Job_Duration
+      expr: (avg(spark_mesos_cluster_driver_launch_to_finish_state_finished_mean) / 60000) >= 75
+      for: 3m
+      labels:
+        severity: major
+      annotations:
+        description: 'Spark jobs taking longer to complete than expected'
+
+    - alert: Spark_Average_Launch_Time
+      expr: (avg(spark_mesos_cluster_executor_start_to_all_launched_mean) / 1000) >= 30
```
elezar: The product requirements mention 99th percentiles. Why are we calculating averages here? Note that this seems to be the average of the mean.

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+    - alert: Spark_Streaming_Jobs
+      expr: count(spark_mesos_cluster_executor_count{task_name=~".*Kafka.*"}) >= 250
+      for: 1m
+      labels:
+        severity: informational
+      annotations:
+        description: 'Spark has reached desired number of Streaming jobs'
+    
+    - alert: Spark_Dispatcher_Queue_Depth
+      expr: spark_mesos_cluster_driver_waiting > 200
```
elezar: Is this a global value? Does each dispatcher not have its own queue?

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+      for: 3m
+      labels:
+        severity: major
+      annotations:
+        description: 'Spark dispatcher queue depth is beyond desired limits'
+
+    - alert: Spark_Average_Job_Duration
+      expr: (avg(spark_mesos_cluster_driver_launch_to_finish_state_finished_mean) / 60000) >= 75
```
elezar: Same question about the use of average?

monitoring/alerts/spark-rules.yml
```diff
@@ -0,0 +1,98 @@
...
+    - alert: Spark_Dispatcher_Count
+      expr: count(spark_mesos_cluster_driver_launched) >= 50
+      for: 1m
+      labels:
+        severity: informational
+      annotations:
+        description: 'Spark has reached desired number of dispatchers'
+
+    - alert: Spark_Dispatcher_Count_Decreasing
+      expr: count(spark_mesos_cluster_driver_launched offset 5m) - count(spark_mesos_cluster_driver_launched) > 0
```
elezar: Would using the 5m rate here not make more sense?

monitoring/alerts/spark-rules.yml
samvantran: Hm, how do we know when we fail to reach that value? It could be failing throughout the entirety of the test until the very end. Are you suggesting something similar to what Marcin's doing with reporting the tests in a `steady` state?

Or would you be okay with something a little more simple like
```
alert: Spark_Job_Failure_Rate
   expr: (sum(spark_mesos_cluste... failed))) < 99.0
   for: 1h
```

monitoring/alerts/spark-rules.yml
samvantran: Honestly, this is what the product requirement is: 
> MUST have a low standard deviation

so your guess is as good as mine in terms of what should be done. It was just something that we wanted alerts on. 

And yes, you're right, stddev =/= variance so I can change the wording but I don't think it will help much. If this alert in general is unclear or unhelpful, maybe we can get rid of it? Seems like product set an arbitrary requirement for this one.

monitoring/alerts/spark-rules.yml
samvantran: Yes, every dispatcher has its own queue so if any reach 200 this alert fires but this is actually a bad expression. This would cause prom to process way more values than it should. Proper alert should be: 
```
expr: count(spark_mesos_cluster_driver_waiting > 200) > 0
``` 
so we count if _any_ dispatcher has a depth greater than 200

monitoring/alerts/spark-rules.yml
elezar: Sorry. My bad.

monitoring/alerts/spark-rules.yml
elezar: We should be alerting when the rate is lower than the desired threshold. This way one could investigate and do something about it. If it is failing during the entire test then we probably have to rethink how we define this. For example, if no jobs are running, we have a 100% success rate. If one job then starts running, we still have a 100% success rate until it fails, at which point we have a 0% success rate and the alert should trigger.

monitoring/alerts/spark-rules.yml
porridge: But yes, `rate()` has goodies like extrapolation, automatic reset recovery, etc.

monitoring/alerts/spark-rules.yml
samvantran: okay i'll change the alert to mark it as a failure if below 99.9%

monitoring/alerts/spark-rules.yml
samvantran: Changing to `rate()`

monitoring/alerts/spark-rules.yml
```diff
@@ -89,10 +89,10 @@ groups:
       annotations:
         description: 'Spark job's launch rate has unusually high standard deviation'
 
-    - alert: Spark_Job_Success_Rate
-      expr: (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) / (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_lost) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_failed))) >= 99.0
+    - alert: Spark_Jobs_Have_Failed
+      expr: (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) / (sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_finished) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_lost) + sum(spark_mesos_cluster_driver_finished_count_mesos_state_task_failed))) < 0.999
```
elezar: Was the product requirement 99.9%? I thought it was 99%?

samvantran: yeah, I checked back and it was 
> MUST have 99.9% of Spark jobs succeed the first time

I believe this is the most up to date ticket w/ all the PM requirements: https://jira.mesosphere.com/browse/DCOS-28030

PR #384, opened 2018-07-18
[COPS-3443] Make default docker image the one in resource.json

elezar:
Specifying the default docker image twice means that the default installation doesn't work on local universes.

This change removes the default from `config.json` which makes the value specified in `resource.json` the default. This does not change behaviour, but makes it easier to use Spark from a local universe.

The alternative would be extensive documentation changes.

### DIFF
--- a/universe/config.json
+++ b/universe/config.json
@@ -44,8 +44,8 @@
                 },
                 "docker-image": {
                     "type": "string",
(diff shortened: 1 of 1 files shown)


mpereira:
Is [`{{ docker-image }}`](https://github.com/mesosphere/spark-build/blob/fef00f7931f096b49e5a97a9bf9aa1f068d603de/universe/resource.json#L5) in `resource.json` interpolated to a value that is compatible with local universes?

mpereira:
Gotcha.

PR #388, opened 2018-08-02
[DCOS-38138] Update Spark CLI for shell-escape fix

samvantran:
Resolves [DCOS-38138](https://jira.mesosphere.com/browse/DCOS-38138) and [DCOS-38462](https://jira.mesosphere.com/browse/DCOS-38462)

This PR builds off [mesosphere/spark#33](https://github.com/mesosphere/spark/pull/33) and updates the Spark CLI to make it possible to pass multiple --conf args in a single string  eg. 
```
--conf spark.driver.extraJavaOptions='-Dspecial.char.param=\$PATH -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam1=\"This setting has a space\"'
```
Previously, only one value was accepted in a typical `--conf arg=val1` and would fail if there were spaces included.

An example spark run command: 
```
dcos spark run --submit-args=" \
	--conf spark.driver.cores=1.0 \
	--conf spark.driver.memory=1g \
	--conf spark.executor.extraJavaOptions='-XX:+PrintGCDetails -XX:+PrintGCTimeStamps' \
	--conf spark.mesos.containerizer=mesos  \
	--class org.apache.spark.examples.SparkPi \
	https://s3-us-west-2.amazonaws.com/infinity-artifacts/spark/spark-examples_2.11-2.2.1.jar 100"
```


manifest.json
```diff
@@ -1,8 +1,8 @@
 {
     "spark_version": "2.2.1",
     "default_spark_dist": {
-        "hadoop_version": "2.6",
-        "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.2.1-2-bin-2.6.tgz"
+        "hadoop_version": "2.7",
+        "uri": "svt-dev.s3.amazonaws.com/spark/spark-2.2.1-bin-2.7.3-dcos-38138.tgz"
```
samvantran: Once PR is in good shape and approved, I'll update the `downloads.mesosphere` uri to include the updated tgz

tests/test_spark.py
```diff
@@ -105,6 +105,18 @@ def test_sparkPi(service_name=utils.SPARK_SERVICE_NAME):
...
 
 
+@pytest.mark.sanity
+def test_spark_with_multi_configs(service_name=utils.SPARK_SERVICE_NAME):
+    utils.run_tests(
+        app_url=utils.dcos_test_jar_url(),
+        app_args="",
+        expected_output="spark.executor.extraJavaOptions,-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3",
+        service_name=service_name,
+        args=["--conf spark.driver.extraJavaOptions='-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3'",
```
samvantran: This is giving me trouble. This same command works via Spark CLI but not with pytests

elezar: Looking at the TeamCity output, the error seems to be thrown by the `Parse` function called here: https://github.com/mesosphere/spark-build/blob/DCOS-38138-shell-escape/cli/dcos-spark/submit_builder.go#L411. It might be a good idea to check whether that is behaving as expected.

Also, confirming that the following command:
```bash
dcos spark --name=spark run --verbose --submit-args="--conf spark.driver.extraJavaOptions='-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3' --conf spark.mesos.containerizer=mesos --class MultiConfs --conf spark.mesos.role=* --conf spark.driver.memory=2g http://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/spark/test-20180802-062841-eHc7mdB3GD1WgGtj/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar "
```
exactly as it is generated by the tests works from the CLI would also be good.

cli/dcos-spark/submit_builder.go
```diff
@@ -280,75 +284,61 @@ func parseApplicationFile(args *sparkArgs) error {
...
+			// if current is a boolean flag, no merge required; eg --supervise
 			for _, boolVal := range boolVals {
-				if boolVal.flagName == arg[2:] {
-					argsEquals = append(argsEquals, arg)
-					i += 1
-					continue ARGLOOP
+				if boolVal.flagName == current[2:] {
+					sparkArgs = append(sparkArgs, current)
+					i++
+					continue LOOP
```
elezar: What if we pass `--supervise false` or something similar? The code doesn't seem to handle this case explicitly (It seems as if this will be joined in the `default` case though.

cli/dcos-spark/submit_builder_test.go
```diff
@@ -49,6 +51,40 @@ func (suite *CliTestSuite) TestCleanUpSubmitArgs() {
 	}
 }
 
+func (suite *CliTestSuite) TestCleanUpSubmitArgsWithMulitpleValues() {
+	_, args := sparkSubmitArgSetup()
+	inputArgs := "--conf spark.driver.extraJavaOptions='-XX:+PrintGC -Dparam1=val1 -Dparam2=val2' main.py 100"
+	expected := "--conf=spark.driver.extraJavaOptions=-XX:+PrintGC -Dparam1=val1 -Dparam2=val2"
```
elezar: Here is the problem. The quoting is stripped from the argument. In the CI, it is most likely the second `-XX` that is failing, but I would expect the `-D` to fail here aswell.

elezar: Do confirm, we may want to include tests that call `buildSubmitJson` with the relevant arguments too.

tests/test_spark.py
samvantran: I can confirm that the exact same command works when submitted via command line and the stdout returns the expected args: 
from CLI: 
```
Translated spark-submit arguments: '--conf=spark.driver.extraJavaOptions=-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3, --conf=spark.mesos.containerizer=mesos, --class=MultiConfs, --conf=spark.mesos.role=*, --conf=spark.driver.memory=2g, http://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/spark/test-20180802-062841-eHc7mdB3GD1WgGtj/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar'
...
Run job succeeded. Submission id: driver-20180802152929-0002
```
stdout
```
1.448: [GC (Allocation Failure) [PSYoungGen: 64512K->8448K(75264K)] 64512K->8456K(247296K), 0.0073157 secs] [Times: user=0.01 sys=0.00, real=0.01 secs] 
Running MultiConfs app. Printing out all config values:
(spark.driver.extraJavaOptions,-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3)
(spark.jars,file:/mnt/mesos/sandbox/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar)
...
```
The issue seems to be going through CI requires another layer of Python string parsing which causes the Go parsing to fail (my guess is that additional quotes are added)

cli/dcos-spark/submit_builder.go
samvantran: You're right, `false` will get joined with `--supervise` and the job will fail. I just tried it and we get a parse error 
```
Error when parsing --submit-args: unknown long flag '--supervise false'
```
In this case, it might make sense to check for boolean flags and if found, check the next argument. If it's a boolean statement like `true/false`, we could just move the pointer forward and skip it (`i+=2`). 

More generally though, there are a lot of ways to mess up confs. Does it make sense to have some kind of blacklist that guards against 'user error'?

cli/dcos-spark/submit_builder_test.go
samvantran: yeah i've tested it on CI where even just setting `extraJavaOptions='val1'` results in error `Cannot read file val1` or something like that. 

I think this is Python doing its own kind of parsing before getting to Go parsing and that causes the issue.

In commit [7c294c7](https://github.com/mesosphere/spark-build/pull/388/commits/7c294c7e19bd6249e5f01bbb7723c4a0c884dd43) I added that same config passed to `buildSubmitJson` and it still works.

samvantran:
@elezar Thanks for taking a look! Yes, the quotes are dropped from the multi-args which works when it gets to Kingpin parsing.

The problem is CI tests use Python and there are a bunch of places where `"{}.format()"` are called which is adding another layer of string parsing. This makes it tricky to get parsing correct across two different languages. My concern is trying to 'make it work' for Python CI when most cases will involve users submitting via DCOS Spark CLI and Python isn't involved.

samvantran:
Hm, surprisingly when I run `test.sh` locally (connected to a personal CCM cluster), the test passes! Maybe it isn't a Python issue... Make/Teamcity? 😞 

run test:
```
╭─  ~/Dev/mesosphere/spark-build                                                                                                                                                           07:42    DCOS-38138-shell-escape    8   14:26:45
╰ CLUSTER_URL=$(dcos config show core.dcos_url) \
STUB_UNIVERSE_URL=https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/spark/20180801-170442-aNDU6gcD0WLsTAVY/stub-universe-spark.json,https://universe-converter.mesosphere.com/transform?url=https://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/spark-history/20180801-015849-bYYlp28dxraqW7Bc/stub-universe-spark-history.json \
 TEST_SH_DCOS_SPARK_TEST_JAR_URL=http://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/spark/test-20180802-172502-iseM14lZPoG40Htm/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar \
 TEST_SH_MESOS_SPARK_TEST_JAR_URL=http://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/spark/test-20180802-202846-cKhPCqE64FKeTtMr/mesos-spark-integration-tests-assembly-0.1.0.jar \
 PYTEST_ARGS='-k test_spark_with_multi_configs' \
 ./test.sh 
```
test output:
```
[2018-08-02 21:08:13,242|sdk_cmd|INFO]: STDOUT:
1.375: [GC (Allocation Failure) [PSYoungGen: 64512K->7368K(75264K)] 64512K->7376K(247296K), 0.0061293 secs] [Times: user=0.01 sys=0.00, real=0.01 secs]
Running MultiConfs app. Printing out all config values:
(spark.driver.extraJavaOptions,-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3)
(spark.jars,file:/mnt/mesos/sandbox/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar)
...
PASSED
==========
======= END: test_spark_py::test_spark_with_multi_configs
==========
```

cli/dcos-spark/submit_builder.go
```diff
@@ -280,75 +284,61 @@ func parseApplicationFile(args *sparkArgs) error {
...
+			next := args[i+1]
+			sparkArgs = append(sparkArgs, current+"="+next)
 			i += 2
-		} else {
-			// already joined or at the end, pass through:
-			argsEquals = append(argsEquals, arg)
-			i += 1
+		default:
+			// if not a flag or jar, current is a continuation of the last arg and should not have been split
+			// eg extraJavaOptions="-Dparam1 -Dparam2" was parsed as [extraJavaOptions, -Dparam1, -Dparam2]
```
susanxhuynh: Oh, this isn't handled by shellwords? (It sounded like it might based on the README.)

tests/test_spark.py
susanxhuynh: Could you add a few special characters here as well? Maybe a `-Dparam3=\"val3 val4\"` like in the unit test?

tests/jobs/scala/src/main/scala/MultiConfs.scala
```diff
@@ -0,0 +1,16 @@
...
+  * Specifically we want to test whether long strings of multiple arg=values are passed through correctly.
+  * E.g. 
+  * dcos spark run --submit-args="--conf=spark.driver.extraJavaOptions='val1 val2 val3' --class MultiConfs"
+  */ 
+object MultiConfs {
+    def main(args: Array[String]): Unit = {
+        println("Running MultiConfs app. Printing out all config values:")
+        val APPNAME = "MultiConfs App"
+        val conf = new SparkConf().setAppName(APPNAME)
+        conf.getAll.foreach(println)
```
susanxhuynh: One additional check that tests the whole thing end-to-end: pass in `-Dparam1=\"valA valB\"` as one of the extraJavaOptions, and then inside this app, you can check the value of System.getProperty("param1") to make sure it was set correctly.

cli/dcos-spark/submit_builder_test.go
susanxhuynh: BTW, I might have missed it in the cleanupSubmitArgs logic: when do the single quotes get stripped off?

samvantran:
I think this may be the culprit: [sdk_cmd.py#L138](https://github.com/mesosphere/spark-build/blob/master/testing/sdk_cmd.py#L138)

Here's what `Translated spark-submit arguments` is **supposed** to look like: 
```
Translated spark-submit arguments: '--conf=spark.driver.extraJavaOptions=-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3, --conf=spark.mesos.containerizer=mesos, --class=MultiConfs, --conf=spark.mesos.role=*, --conf=spark.driver.memory=2g, http://infinity-artifacts-ci.s3.amazonaws.com/autodelete7d/spark/test-20180802-202846-cKhPCqE64FKeTtMr/dcos-spark-scala-tests-assembly-0.2-SNAPSHOT.jar'
```
> a comma separated list of strings (due to this [line](https://github.com/mesosphere/spark-build/pull/388/files?utf8=%E2%9C%93&diff=unified#diff-4b81b743bcae14a20623a9e94b14a92bR338))

Here's what the arguments look like in [the CI job](https://teamcity.mesosphere.io/viewLog.html?buildId=1150073&buildTypeId=DcosIo_DcosCommons_Spark_SparkBuildPr&tab=buildLog&_focus=23942): 
```
[Step 7/8] 2018/08/02 21:10:51 Translated spark-submit arguments: '[--conf=spark.driver.extraJavaOptions='-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -Dparam3=val3' ....jar]'
```
> it's one string encased inside brackets `[]`

Interestingly, two lines up in the CI build log, you see that the dcos command about to be run looks normal: 
```
[Step 7/8] sdk_cmd.py  137 INFO     (CLI) dcos spark --name=spark run --verbose --submit-args="--conf spark.driver.extraJavaOptions='-XX:+PrintGCDetails... 
```
which means somewhere in [this call](https://github.com/mesosphere/spark-build/blob/master/testing/sdk_cmd.py#L138), something goes wrong.

I did some digging and it looks like `subprocess.run(..., shell=True)` is **strongly discouraged** (see [here](https://docs.python.org/2/library/subprocess.html#frequently-used-arguments)). Besides it being dangerous, it also means that whichever `/bin/sh` exists on the machine is the subprocess created and we are at its mercy.

Recommended approaches are to instead `subprocess.run()` with an array of strings and set `shell=False`, or escape strings using `shlex.quote()` ([link](https://docs.python.org/3.4/library/shlex.html#shlex.quote)).

That means making this Spark CLI fix work in CI requires making it work through `Python -> Shell -> Go` and that is _very cumbersome_ 😢 

Maybe calling `dcos spark` via a `dcos task exec` might be easier. @elezar @susanxhuynh what do you think?

cli/dcos-spark/submit_builder.go
```diff
@@ -280,75 +284,61 @@ func parseApplicationFile(args *sparkArgs) error {
 	return nil
 }
 
+// we use Kingpin to parse CLI commands and options
+// spark-submit by convention uses '--arg val' while kingpin only supports --arg=val
+// cleanUpSubmitArgs transforms the former to the latter
 func cleanUpSubmitArgs(argsStr string, boolVals []*sparkVal) ([]string, []string) {
```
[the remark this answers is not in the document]
samvantran: Hm, how about renaming to `transformSubmitArgs?` ?

cli/dcos-spark/submit_builder_test.go
samvantran: Not in our logic, it happens in `shellwords.parse()` 
```
args, err := shellwords.Parse(argsStr)
```

cli/dcos-spark/submit_builder.go
```diff
@@ -280,75 +284,61 @@ func parseApplicationFile(args *sparkArgs) error {
...
+	if err != nil {
+		log.Fatalf("Could not parse string args correctly. Error: %v", err)
+	}
+	sparkArgs, appArgs := make([]string, 0), make([]string, 0)
+LOOP:
+	for i := 0; i < len(args); {
+		current := strings.TrimSpace(args[i])
+		switch {
+		// if current is a spark jar/app, we've processed all flags; add jar to sparkArgs and append the rest to appArgs
+		case strings.HasSuffix(current, ".jar") || strings.HasSuffix(current, ".r") || strings.HasSuffix(current, ".py"):
```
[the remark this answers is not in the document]
samvantran: Good catch! It probably does matter so I'll change it back

cli/dcos-spark/submit_builder.go
samvantran: No unfortunately the library splits the string on non-escaped spaces so this `default` case glues the string back together.

tests/jobs/scala/src/main/scala/MultiConfs.scala
samvantran: Good idea, I'll update.

cli/dcos-spark/submit_builder.go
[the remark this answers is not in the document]
elezar: In fact. How about having an `isAppArtifact` or function or something? (with unit tests to ensure that we don't subtly change the behaviour)

spark-testing/spark_utils.py
```diff
@@ -147,14 +148,26 @@ def submit_job(
 
     if write_conf_to_temp_file:
         args_file = tempfile.NamedTemporaryFile("w")
-        args_file.write(submit_args)
+        args_file.write('dcos spark run {} --submit-args="{}"'.format(verbose_flag, submit_args))
```
susanxhuynh: If you write the entire command to a file, you can just execute the file as a shell script: like
`subprocess.run(args_file.name)`
rather than cat-ing the contents of the files.

spark-testing/spark_utils.py
```diff
@@ -140,11 +145,29 @@ def submit_job(
 
     submit_args = ' '.join([' '.join(conf_args), app_url, app_args])
     verbose_flag = "--verbose" if verbose else ""
-    stdout = sdk_cmd.svc_cli(
-        SPARK_PACKAGE_NAME,
-        service_name,
-        'run {} --submit-args="{}"'.format(verbose_flag, submit_args))
-    result = re.search(r"Submission id: (\S+)", stdout)
+
+    if write_conf_to_temp_file:
```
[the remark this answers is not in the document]
samvantran: this felt like the smallest change to accommodate writing the entire command to a file like others have suggested. If you prefer, I could create a new function `submit_job_via_file` that mostly duplicates `submit_job` if you prefer I leave the original alone. 

Or perhaps you have something else in mind?

tests/jobs/scala/src/main/scala/MultiConfs.scala
```diff
@@ -0,0 +1,31 @@
...
+        // Verify property is set in Spark conf
+        val conf = new SparkConf().setAppName(AppName)
+        println("Printing all conf values...")
+        conf.getAll.foreach(println)
+        
+        // Verify property is set in system 
+        val props = System.getProperties()
+        props.list(System.out)
+        val key = "param3"
+        val expectedVal = "valA valB"
```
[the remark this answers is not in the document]
samvantran: Susan asked to check System.props to verify that values were passed end to end but I took that literally. Originally I was printing all values so I agree with you, that's a better route.

cli/dcos-spark/submit_builder.go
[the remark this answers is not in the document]
samvantran: If this is true, it will be a major 
getinsertpic.com

cli/dcos-spark/submit_builder.go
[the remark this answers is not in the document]
samvantran: done [6b32e5](https://github.com/mesosphere/spark-build/pull/388/commits/6b32e56d42cb7e7667d270acbabecd31641a950a)

samvantran:
@elezar @susanxhuynh hallelujah! CI finally passes. Please review hopefully one last time 🙏 

Quick recap: 

1. CI was misconfigured for the longest time. Over the last few months, only the latest spark package in universe was being tested, not any actual stubs. This explains why the 20 commits I made trying different strategies to make CI happy always failed.
2. The upstream shell-escape PR actually needed fixing which I did in [this PR](https://github.com/mesosphere/spark/pull/33). It works for our fork but eventually will need an upstream fix to apache/spark.
3. I did have to comment out one test in commit [7865b6c](https://github.com/mesosphere/spark-build/pull/388/commits/7865b6c1603d719fcc35508b2d61eff8aaded87c) for reasons I explained in [this ticket](https://jira.mesosphere.com/browse/DCOS-41088)

test.sh
```diff
@@ -10,12 +10,13 @@
 # Exit immediately on errors
 set -e
 
+timestamp="$(date +%y%m%d-%H%M%S)"
```
samvantran: PR #398 covers this file change

cli/dcos-spark/submit_builder.go
```diff
@@ -280,75 +292,87 @@ func parseApplicationFile(args *sparkArgs) error {
 	return nil
 }
 
-func cleanUpSubmitArgs(argsStr string, boolVals []*sparkVal) ([]string, []string) {
-
-	// collapse two or more spaces to one.
-	argsCompacted := collapseSpacesPattern.ReplaceAllString(argsStr, " ")
+// we use Kingpin to parse CLI commands and options
+// spark-submit by convention uses '--arg val' while kingpin only supports --arg=val
```
elezar: Could we create a ticket to address this. Kingpin supports *both* `--arg val` and `--arg=val`.

cli/dcos-spark/submit_builder.go
```diff
@@ -280,75 +292,87 @@ func parseApplicationFile(args *sparkArgs) error {
...
-	return argsEquals, appFlags
+func checkIfBooleanFlag(boolVals []*sparkVal, args []string, idx int) (isBool, skipNext bool) {
+	// check if current is a boolean flag (eg --supervise) but skip any accompanying boolean values (eg --supervise true)
+	current := args[idx]
+	for _, boolVal := range boolVals {
+		if boolVal.flagName == current[2:] {
+			isBool = true
+			next := args[idx+1]
+			if _, err := strconv.ParseBool(next); err == nil {
+				skipNext = true
```
elezar: This is not correct. What if someone calls the function with `--supervise false` that should have different behaviour than `--supervise true`, but this is not captured here.

Also, as far as I know named return values have fallen out of favour in idiomatic Go.

cli/dcos-spark/submit_builder_test.go
```diff
@@ -49,12 +51,89 @@ func (suite *CliTestSuite) TestCleanUpSubmitArgs() {
...
+func (suite *CliTestSuite) TestCheckIfBooleanFlagWithBooleanValue() {
+	_, args := sparkSubmitArgSetup()
+	idx := 0
+
+	inputArgs := []string{"--supervise", "true", "--conf", "spark.mesos.containerizer=mesos"}
+	isBool, skipNext := checkIfBooleanFlag(args.boolVals, inputArgs, idx)
+	assert.True(suite.T(), isBool, "isBool should be true since --supervise is a Boolean flag")
+	assert.True(suite.T(), skipNext, "skipNext should be true since 'true' is a Boolean value")
+
+	inputArgs = []string{"--supervise", "false", "--conf", "spark.mesos.containerizer=mesos"}
```
elezar: What about `--supervise` with NO args?

I also don't see a test for the equivalence of:
* `--supervise` and `--supervise true`
* (missing `--supervise`) and `--supervise false`

cli/dcos-spark/submit_builder.go
samvantran: here you go: https://jira.mesosphere.com/browse/DCOS-41107

cli/dcos-spark/submit_builder.go
samvantran: Ah... that's true. It may make sense to remove it from the list in the `false` case since that is default Spark behavior.

As for point 2, I didn't know. I was just operating off the Go reference and thought it was a nifty feature.

I did some googling and found this as the latest reference but it was in favor of named return values: https://blog.minio.io/golang-internals-part-2-nice-benefits-of-named-return-values-1e95305c8687

Can you point me to resources that argue the opposite?

manifest.json
```diff
@@ -2,7 +2,7 @@
     "spark_version": "2.2.1",
     "default_spark_dist": {
         "hadoop_version": "2.6",
-        "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.2.1-2-bin-2.6.tgz"
+        "uri": "svt-dev.s3.amazonaws.com/spark/spark-2.2.1-bin-2.6.5-DCOS-38138.tgz"
```
susanxhuynh: I guess this will be updated once the next release is cut?

samvantran: Yup, or do you think I should first merge the [mesosphere/spark PR](https://github.com/mesosphere/spark/pull/33) (and revert the --proxy-user commit) off `custom-branch-2.2.1-X`? I could then create a new tgz and upload it to 
`downloads.mesosphere.com/spark/assets/spark-2.2.1-3-bin-2.6.tgz` (note the `-3`) and change this PR. That way I wouldn't have to do it twice?

susanxhuynh: Hm, I don't think we should keep overwriting the "2.2.1-3" tgz with different versions. Until we actually make a release, how about something like "2.2.1-3-SNAPSHOT" to signal that it's not an officially released version, but a dev one that will keep changing up until the release.

samvantran: SNAPSHOT sounds best

cli/dcos-spark/submit_builder.go
elezar: Great! Thanks.

cli/dcos-spark/submit_builder.go
elezar: I may have just been recalling a recording of a gophercon talk that I watched. In this case, I don't think it's the named return values so much as the naked return that is the issue. I stand corrected.

Here is a PR review that makes the changes to the Go standard library and style guide: 
* https://go-review.googlesource.com/c/go/+/20024 
* https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

For clarity here:
> Named returned values should only be used on public funcs and methods
> when it contributes to the documentation.
>
> Named return values should not be used if they're only saving the
> programmer a few lines of code inside the body of the function,
> especially if that means there's stutter in the documentation or it
> was only there so the programmer could use a naked return
> statement. (Naked returns should not be used except in very small
> functions)

in this case, I would say that we are not stuttering so the named values themselves are ok, but we should consider dropping the naked return -- although we could consider this a very small function.

cli/dcos-spark/submit_builder_test.go
elezar: I'm all for not supporting `--bool-flag BOOL_VAL` options, but then we need to change the handling of the bool args. The thing to confirm would then be that we correctly handle the different cases:
* `--supervise` should pass
* `--supervise true` should error
* `--supervise false` should error
* `--supervise https://some.jar` should pass

Are these all the test cases that we need to cover?

cli/dcos-spark/submit_builder.go
```diff
@@ -280,75 +292,87 @@ func parseApplicationFile(args *sparkArgs) error {
...
-	client.PrintVerbose("Translated spark-submit arguments: '%s'", argsEquals)
-	client.PrintVerbose("Translated application arguments: '%s'", appFlags)
+	return false
+}
 
-	return argsEquals, appFlags
+func checkIfBooleanFlag(boolVals []*sparkVal, args []string, idx int) (isBool, skipNext bool) {
+	// check if current is a boolean flag (eg --supervise) but skip any accompanying boolean values (eg --supervise true)
+	current := args[idx]
+	for _, boolVal := range boolVals {
```
elezar: Looking at this again this morning. We should break out of this loop as soon as we have found the matching flag -- either by returning, or by calling `break`.

cli/dcos-spark/submit_builder.go
```diff
@@ -6,20 +6,30 @@ import (
...
+	"github.com/mesosphere/dcos-commons/cli/config"
+	"gopkg.in/alecthomas/kingpin.v3-unstable"
 )
 
 var keyWhitespaceValPattern = regexp.MustCompile("(.+)\\s+(.+)")
 var backslashNewlinePattern = regexp.MustCompile("\\s*\\\\s*\\n\\s+")
 var collapseSpacesPattern = regexp.MustCompile(`[\s\p{Zs}]{2,}`)
 
+// AcceptedSparkAppExtensions lists the accepted extensions for Spark application artifacts
+var AcceptedSparkAppExtensions = []string{
```
elezar: Does this have to be public? Also, we could make the list local to the only function where it is used.

cli/dcos-spark/submit_builder.go
samvantran: here you go: [d0aae3c](https://github.com/mesosphere/spark-build/pull/388/commits/d0aae3caab69415604edf950a42b4e1a57517fe0)

PR #391, opened 2018-08-13
Add comments to configs for 08-08 tests

elezar:
This is a quick PR to add comments for the configs used in the 08-08 scale test. My arithmetic was off and we set the number of words incorrectly for the finite jobs (making them effectively infinite)


elezar:
> Thanks @elezar. Just curious: why not 7200 instead of 7692?

It could be that the weekend was too long, but when I calculated it 7692 gave the same runtime as the previous test:
```
In [2]: (100000 / 13 / 60 / 60) * 60 * 60
Out[2]: 7692.307692307692
```
(I should have just divided 100000 by 13 instead of multiplying)

We could follow up and rather specify the duration directly and calculate the number of words instead?

mpereira:
@elezar gotcha.

> We could follow up and rather specify the duration directly and calculate the number of words instead?

Yeah, maybe that makes more sense. The idea of specifying the number of words was to specify (roughly) the total amount of bytes to be produced.

PR #392, opened 2018-08-14
[DCOS-39050] Added files for Hive Docker image

susanxhuynh:
This Hive Docker image is intended for use in Spark integration tests.

It's based on https://github.com/tilakpatidar/cdh5_hive_postgres (the one Evan found) with the following changes:

- Removed docker-compose (not supported by Marathon)
- Added Kerberos (see the `kerberos/` directory)
    * Kerberized versions of hadoop, hive xml config files
    * krb5.conf
    * Marathon config to run the container
- Hive: changes in `bootstrap.sh` for Kerberos
- You can also refer to [my fork](https://github.com/susanxhuynh/cdh5_hive_postgres/commits/master) to see what I added.


susanxhuynh:
Tests are passing now. The failures yesterday seem to have been temporary connectivity problems at the sbt mirror site.

tools/hive/ubuntu/base.env
```diff
@@ -0,0 +1,19 @@
...
+POSTGRESQL_CONFIG_FILE=/var/lib/postgresql/9.5/main/postgresql.conf
+POSTGRESQL_BIN=/usr/lib/postgresql/9.5/bin/postgres
+PGPASSWORD=hive
+HADOOP_HOME=/usr/local/hadoop
+HADOOP_PREFIX=/usr/local/hadoop
+HADOOP_COMMON_HOME=/usr/local/hadoop
+HADOOP_HDFS_HOME=/usr/local/hadoop
+HADOOP_MAPRED_HOME=/usr/local/hadoop
+HADOOP_YARN_HOME=/usr/local/hadoop
+HADOOP_CONF_DIR=/usr/local/hadoop/etc/hadoop
```
samvantran: Needs a newline?

tools/hive/ubuntu/base.env
```diff
@@ -0,0 +1,19 @@
...
+POSTGRESQL_MAIN=/var/lib/postgresql/9.5/main/
+POSTGRESQL_CONFIG_FILE=/var/lib/postgresql/9.5/main/postgresql.conf
+POSTGRESQL_BIN=/usr/lib/postgresql/9.5/bin/postgres
+PGPASSWORD=hive
+HADOOP_HOME=/usr/local/hadoop
+HADOOP_PREFIX=/usr/local/hadoop
+HADOOP_COMMON_HOME=/usr/local/hadoop
+HADOOP_HDFS_HOME=/usr/local/hadoop
+HADOOP_MAPRED_HOME=/usr/local/hadoop
+HADOOP_YARN_HOME=/usr/local/hadoop
```
samvantran: Are all of these supposed to point to the same dir?

tools/hive/hadoop-2.6.0/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ENV HADOOP_MAPRED_HOME /usr/local/hadoop
+ENV HADOOP_YARN_HOME /usr/local/hadoop
+ENV HADOOP_CONF_DIR /usr/local/hadoop/etc/hadoop
+
+ENV PATH $PATH:$JAVA_HOME/bin:$HIVE_HOME/bin:$HADOOP_HOME:$HADOOP_HOME/bin
+
+ADD ./deps/hadoop-${HADOOP_VERSION}-cdh${CDH_EXACT_VERSION}.tar.gz /usr/local
+RUN cd /usr/local && ln -s ./hadoop-${HADOOP_VERSION}-cdh${CDH_EXACT_VERSION} hadoop
+
+RUN sed -i '/^export JAVA_HOME/ s:.*:export JAVA_HOME=/usr/lib/jvm/java-8-oracle\nexport HADOOP_PREFIX=/usr/local/hadoop\nexport HADOOP_HOME=/usr/local/hadoop\n:' $HADOOP_PREFIX/etc/hadoop/hadoop-env.sh
```
elezar: This seems to be explicitly setting the contents of the `/etc/hadoop/hadoop-env.sh` file. Why is it not sufficient to rely on the environment variables in this case?

tools/hive/hadoop-2.6.0/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN sed -i '/^export HADOOP_CONF_DIR/ s:.*:export HADOOP_CONF_DIR=/usr/local/hadoop/etc/hadoop/:' $HADOOP_PREFIX/etc/hadoop/hadoop-env.sh
+
+# copy hadoop site xml files
+RUN mkdir $HADOOP_PREFIX/input
```
elezar: Is this just to preserve the contents of the original files?

tools/hive/hadoop-2.6.0/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ADD templates/core-site.xml.template $HADOOP_PREFIX/etc/hadoop/core-site.xml.template
+ADD templates/hdfs-site.xml.template $HADOOP_PREFIX/etc/hadoop/hdfs-site.xml.template
+ADD conf/mapred-site.xml $HADOOP_PREFIX/etc/hadoop/mapred-site.xml
+ADD templates/yarn-site.xml.template $HADOOP_PREFIX/etc/hadoop/yarn-site.xml.template
+
+# format namenode
+RUN $HADOOP_PREFIX/bin/hdfs namenode -format
+
+# fixing the libhadoop.so
+RUN rm -rf /usr/local/hadoop/lib/native/*
```
elezar: Why is it required to fix the native libraries?

tools/hive/hadoop-2.6.0/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+RUN sed -i '/^export JAVA_HOME/ s:.*:export JAVA_HOME=/usr/lib/jvm/java-8-oracle\nexport HADOOP_PREFIX=/usr/local/hadoop\nexport HADOOP_HOME=/usr/local/hadoop\n:' $HADOOP_PREFIX/etc/hadoop/hadoop-env.sh
+RUN sed -i '/^export HADOOP_CONF_DIR/ s:.*:export HADOOP_CONF_DIR=/usr/local/hadoop/etc/hadoop/:' $HADOOP_PREFIX/etc/hadoop/hadoop-env.sh
+
+# copy hadoop site xml files
+RUN mkdir $HADOOP_PREFIX/input
+RUN cp $HADOOP_PREFIX/etc/hadoop/*.xml $HADOOP_PREFIX/input
+
+# pseudo distributed configurations
+ADD templates/core-site.xml.template $HADOOP_PREFIX/etc/hadoop/core-site.xml.template
```
elezar: The location of the config files is arbitrary and one should be able to add the generated ones to a specified location and use it as follows: https://github.com/mesosphere/dcos-tensorflow/blob/master/frameworks/tensorflow/src/main/dist/svc.yml#L71

tools/hive/hadoop-2.6.0/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ADD templates/hdfs-site.xml.template $HADOOP_PREFIX/etc/hadoop/hdfs-site.xml.template
+ADD conf/mapred-site.xml $HADOOP_PREFIX/etc/hadoop/mapred-site.xml
+ADD templates/yarn-site.xml.template $HADOOP_PREFIX/etc/hadoop/yarn-site.xml.template
+
+# format namenode
+RUN $HADOOP_PREFIX/bin/hdfs namenode -format
```
elezar: Does it make sense to format the namenode as part of the docker build command?

tools/hive/hadoop-2.6.0/LICENSE
```diff
@@ -0,0 +1,202 @@
+Apache License
```
elezar: Is this license file required? We don't seem to be including any of the hadoop source here.

tools/hive/hadoop-2.6.0/scripts/bootstrap.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -x
+/usr/sbin/sshd
```
elezar: Is there a reason that we have to start an SSH daemon?

tools/hive/hadoop-2.6.0/scripts/bootstrap.sh
```diff
@@ -0,0 +1,24 @@
...
+: ${HADOOP_PREFIX:=/usr/local/hadoop}
+
+$HADOOP_PREFIX/etc/hadoop/hadoop-env.sh
+
+rm /tmp/*.pid
+
+# installing libraries if any - (resource urls added comma separated to the ACP system variable)
+cd $HADOOP_PREFIX/share/hadoop/common ; for cp in ${ACP//,/ }; do  echo == $cp; curl -LO $cp ; done; cd -
```
elezar: We don't seem to use this functionality. We could consider just dropping it.

tools/hive/hadoop-2.6.0/scripts/bootstrap.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# templating of config files
+sed s/HOSTNAME/$HOSTNAME/ /usr/local/hadoop/etc/hadoop/core-site.xml.template > /usr/local/hadoop/etc/hadoop/core-site.xml
```
elezar: It would be great to use mustache templates as we do with our other packages.

tools/hive/hadoop-2.6.0/templates/yarn-site.xml.template
```diff
@@ -0,0 +1,64 @@
+<configuration>
+    <property>
+        <name>yarn.nodemanager.aux-services</name>
+        <value>mapreduce_shuffle</value>
+    </property>
+
+    <property>
+      <name>yarn.application.classpath</name>
```
elezar: From: https://hadoop.apache.org/docs/r2.6.0/hadoop-yarn/hadoop-yarn-common/yarn-default.xml
```
CLASSPATH for YARN applications. A comma-separated list of CLASSPATH entries. When this value is empty, the following default CLASSPATH for YARN applications would be used. For Linux: $HADOOP_CONF_DIR, $HADOOP_COMMON_HOME/share/hadoop/common/*, $HADOOP_COMMON_HOME/share/hadoop/common/lib/*, $HADOOP_HDFS_HOME/share/hadoop/hdfs/*, $HADOOP_HDFS_HOME/share/hadoop/hdfs/lib/*, $HADOOP_YARN_HOME/share/hadoop/yarn/*, $HADOOP_YARN_HOME/share/hadoop/yarn/lib/* 
```
Which seems to cover everything except: `/usr/local/hadoop/etc/hadoop` Is there a specific reason to have to add the single path?

tools/hive/hadoop-2.6.0/templates/yarn-site.xml.template
```diff
@@ -0,0 +1,64 @@
...
+    If enabled, the TimelineClient library used by end-users will post entities
+    and events to the Timeline server.</description>
+    <name>yarn.timeline-service.enabled</name>
+    <value>true</value>
+  </property>
+
+  <property>
+    <description>The hostname of the Timeline service web application.</description>
+    <name>yarn.timeline-service.hostname</name>
+    <value>0.0.0.0</value>
```
elezar: This is already the default value: https://hadoop.apache.org/docs/r2.6.0/hadoop-yarn/hadoop-yarn-common/yarn-default.xml

tools/hive/hive_pg/conf/hive-log4j.properties
```diff
@@ -0,0 +1,88 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
elezar: Was it required to override the defaults?

tools/hive/hive_pg/conf/log4j.properties
```diff
@@ -0,0 +1,6 @@
+log4j.rootLogger=${hive.root.logger}
```
elezar: Was it required to override the defaults?

tools/hive/hive_pg/scripts/bootstrap.sh
```diff
@@ -0,0 +1,49 @@
...
+# create hdfs directories
+$HADOOP_PREFIX/bin/hdfs dfs -mkdir -p /user/root
+hdfs dfs -chown -R hdfs:supergroup /user
+
+$HADOOP_PREFIX/bin/hdfs dfs -mkdir -p /apps/hive/warehouse
+hdfs dfs -chown -R hive:supergroup /apps/hive
+hdfs dfs -chmod 777 /apps/hive/warehouse
+
+# altering the hive-site configuration
+sed s/HOSTNAME/$HOSTNAME/ /usr/local/hive/conf/hive-site.xml.template > /usr/local/hive/conf/hive-site.xml
```
elezar: See previous comment about using mustache templating.

tools/hive/hive_pg/scripts/bootstrap.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -x
+#save all env vars .bashrc for ssh sessions
+printenv | cat >> /root/.bashrc
```
elezar: It shouldn't be required to enable the SSH daemon.

tools/hive/kerberos/templates/hive-site.xml.template
```diff
@@ -0,0 +1,197 @@
+<configuration>
```
elezar: Could we remove anything from this file that's a default value?

tools/hive/kerberos/templates/yarn-site.xml.template
```diff
@@ -0,0 +1,85 @@
+<configuration>
+    <property>
```
elezar: Similar comments as for the non-kerberized version. Switching to mustache should allow us to to unify the templates a little.

tools/hive/ubuntu/Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM ubuntu:trusty
```
elezar: This is pretty old version of ubuntu. Furthermore, all this image seems to do is add the SSD Daemon (which I'm not sure if we need), and set environment variables that are later overwritten.

tools/hive/ubuntu/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+USER root
+
+ENV JAVA_HOME /usr/lib/jvm/java-8-oracle
+ENV HIVE_HOME /usr/local/hive
+ENV HADOOP_HOME /usr/local/hadoop
+
+ENV PATH $PATH:$JAVA_HOME/bin:$HIVE_HOME/bin:$HADOOP_HOME:$HADOOP_HOME/bin
```
elezar: Here we are setting the path to folders that don't exist yet.

tools/hive/ubuntu/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+# install dev tools
+RUN apt-get update
+RUN apt-get install -y curl wget tar openssh-server openssh-client rsync python-software-properties apt-file apache2
+
```
elezar: We're not cleaning up the cache.

tools/hive/ubuntu/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+# for running sshd in ubuntu trusty. https://github.com/docker/docker/issues/5704
+RUN mkdir /var/run/sshd
+RUN echo 'root:secretpasswd' | chpasswd
+RUN sed -i 's/PermitRootLogin without-password/PermitRootLogin yes/' /etc/ssh/sshd_config
+RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
+RUN sed -i 's/Listen 80/Listen 9999/g' /etc/apache2/ports.conf
```
elezar: Is there a reason that we need an Apache webserver too?

tools/hive/hadoop-2.6.0/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ENV JAVA_HOME /usr/lib/jvm/java-8-oracle
+ENV HADOOP_VERSION 2.6.0
+ENV CDH_VERSION 5
+ENV CDH_EXACT_VERSION 5.11.0
+ENV HADOOP_HOME /usr/local/hadoop
+ENV HADOOP_PREFIX /usr/local/hadoop
+ENV HADOOP_COMMON_HOME /usr/local/hadoop
+ENV HADOOP_HDFS_HOME /usr/local/hadoop
+ENV HADOOP_MAPRED_HOME /usr/local/hadoop
+ENV HADOOP_YARN_HOME /usr/local/hadoop
```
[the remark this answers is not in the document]
susanxhuynh: I thought they might be used while the container is running, but let me check on that.

tools/hive/kerberos/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ADD templates/hive-site.xml.template $HIVE_CONF/hive-site.xml.template
+
+# krb5.conf
+ADD conf/krb5.conf /etc/
+
+# install kinit, used in bootstrap script
+RUN apt-get install -y krb5-user
+
+# run bootstrap script
+CMD ["/etc/hive-bootstrap.sh", "-d"]
```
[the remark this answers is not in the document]
susanxhuynh: This command sets up a few directories and then starts the Hive servers. To create and query Hive tables, we would install Spark on the cluster and run a Spark job ... the job would point to the Hive container's IP address. Alternatively, you could go inside the container and start the Hive "beeline" program, in which you could also perform Hive queries.

susanxhuynh:
@elezar 
- I will reduce the number of images.
- I can remove the apache2 web server but not sshd or postgres.
- Sshd is a prerequisite for Hadoop. See https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/SingleCluster.html#Setup_passphraseless_ssh
- Postgres is needed to run Hive in remote mode, which is necessary to reproduce the Spark bug that I need to repro (https://issues.apache.org/jira/browse/SPARK-24338). The exception occurs in the "ThriftHiveMetastore" Client class, which is only used in Hive remote mode. More information about remote mode: https://www.cloudera.com/documentation/enterprise/latest/topics/cdh_ig_hive_metastore_configure.html#topic_18_4_1__title_508
- I'll move this docker image into a separate repo, but I'll make changes here for now since you and Sam have already left comments here.

tools/hive/ubuntu/Dockerfile
susanxhuynh: I don't think this is actually used. I will remove it.

tools/hive/hadoop-2.6.0/Dockerfile
susanxhuynh: That is strange and Hadoop seems to work without this step, with the exception of setting JAVA_HOME. I don't know why, but it seems that JAVA_HOME has to be set directly in this file. https://stackoverflow.com/questions/14325594/working-with-hadoop-localhost-error-java-home-is-not-set

tools/hive/hadoop-2.6.0/Dockerfile
susanxhuynh: The native libraries are not part of the Cloudera distribution. OTOH, hadoop seems to work okay without the native libraries.

tools/hive/hadoop-2.6.0/Dockerfile
susanxhuynh: I guess, removing ...

susanxhuynh:
First pass at combining ubuntu + hadoop + hive into a single image (currently under the "single-image" directory). Still a WIP.

tools/hive/hadoop-2.6.0/Dockerfile
elezar: I think the native libraries are only required when using Hadoop from languages such as C/C++.

tools/hive/hadoop-2.6.0/scripts/bootstrap.sh
susanxhuynh: @elezar That's a good idea. What did you have in mind? I could continue to use `sed` except with the curly braces "{{HOSTNAME}}" in the template. Or, were you thinking of using an actual mustache implementation, such as the Go library in dcos-commons?

tools/hive/hadoop-2.6.0/scripts/bootstrap.sh
susanxhuynh: dropped

tools/hive/hadoop-2.6.0/scripts/bootstrap.sh
susanxhuynh: That is used by Hadoop.

susanxhuynh:
@samvantran @elezar Gentle ping. See comment above.

tools/hive/hadoop-hive/templates/yarn-site.xml.template
```diff
@@ -0,0 +1,2 @@
+<configuration>
+</configuration>
```
samvantran: Is this file necessary? It's essentially empty.

tools/hive/hadoop-hive/scripts/hadoop-bootstrap.sh
```diff
@@ -0,0 +1,27 @@
...
+
+: ${HADOOP_PREFIX:=/usr/local/hadoop}
+
+$HADOOP_PREFIX/etc/hadoop/hadoop-env.sh
+
+rm /tmp/*.pid
+
+# templating of config files
+sed s/{{HOSTNAME}}/$HOSTNAME/ /usr/local/hadoop/etc/hadoop/core-site.xml.template > /usr/local/hadoop/etc/hadoop/core-site.xml
+sed s/{{HOSTNAME}}/$HOSTNAME/ /usr/local/hadoop/etc/hadoop/yarn-site.xml.template > /usr/local/hadoop/etc/hadoop/yarn-site.xml
```
samvantran: Follow up to my question below, it doesn't look like `yarn-site.xml.template` has a `{{HOSTNAME}}` space to fill

tools/hive/hadoop-hive/templates/yarn-site.xml.template
susanxhuynh: Strictly speaking, it's not necessary, but it serves as a placeholder for the "generate_configs.sh", so that the script can treat all config files (including `yarn-site`) equally.

tools/hive/hadoop-hive/scripts/hadoop-bootstrap.sh
susanxhuynh: I'll admit this one is a little confusing. What happens is this script gets called in both the non-kerberized and kerberized images. And, the kerberized version of `yarn-site.xml` does have a `{{HOSTNAME}}` in it, and I wanted to avoid special processing to account for that.

tools/hive/hadoop-hive/scripts/hadoop-bootstrap.sh
```diff
@@ -0,0 +1,27 @@
...
+sed s/{{HOSTNAME}}/$HOSTNAME/ /usr/local/hadoop/etc/hadoop/hdfs-site.xml.template > /usr/local/hadoop/etc/hadoop/hdfs-site.xml
+
+# format namenode
+$HADOOP_PREFIX/bin/hdfs namenode -format
+
+# start hdfs and yarn
+$HADOOP_PREFIX/sbin/start-dfs.sh
+$HADOOP_PREFIX/sbin/start-yarn.sh
+
+if [[ $1 == "-bash" ]]; then
```
elezar: Does this not mean that the container will terminate if the `-bash` argument is not provided?

tools/hive/hadoop-hive/scripts/hive-bootstrap.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -x
+#save all env vars .bashrc for ssh sessions
+printenv | cat >> /root/.bashrc
+
+# hadoop bootstrap
+/etc/hadoop-bootstrap.sh -d
```
elezar: There is no `-d` flag in the `hadoop-bootstrap.sh` script above. Is this intentional?

tools/hive/hadoop-hive/scripts/hive-bootstrap.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# start hive server
+$HIVE_HOME/bin/hive --service hiveserver2 &
+
+
+if [[ $1 == "-bash" ]]; then
+  /bin/bash
+fi
+
+if [[ $1 == "-d" ]]; then
```
elezar: We could use an `elif` here instead followed by an `else` block that prints something if the argument is unknown.

tools/hive/hadoop-hive/scripts/hive-bootstrap.sh
```diff
@@ -0,0 +1,49 @@
...
+hdfs dfs -chmod 777 /apps/hive/warehouse
+
+# altering the hive-site configuration
+sed s/{{HOSTNAME}}/$HOSTNAME/ /usr/local/hive/conf/hive-site.xml.template > /usr/local/hive/conf/hive-site.xml
+sed s/{{HOSTNAME}}/$HOSTNAME/ /opt/files/hive-site.xml.template > /opt/files/hive-site.xml
+
+# start hive metastore server
+$HIVE_HOME/bin/hive --service metastore &
+
+sleep 20
```
elezar: Is there a better way to check for readiness?

tools/hive/kerberos/conf/krb5.conf
```diff
@@ -0,0 +1,10 @@
+[libdefaults]
+default_realm = LOCAL
+dns_lookup_realm = true
+dns_lookup_kdc = true
+udp_preference_limit = 1
+
+[realms]
+LOCAL = {
+  kdc = kdc.marathon.mesos:2500
```
elezar: In our other applications, we use a different endpoint here Should we make this configurable too?

tools/hive/kerberos/scripts/generate_configs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Generate kerberos versions of Hadoop config files by taking the unkerberized versions
+# and tacking on the associated kerberos properties.
+# Assumption is that these two sets of properties do not overlap.
+
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+for FILE_BASE in core-site hdfs-site hive-site yarn-site; do
+    COMBINED_FILE="../templates/${FILE_BASE}.xml.template"
+    echo "Generating config file: kerberos/templates/${FILE_BASE}.xml.template"
```
elezar: Not a blocker, but if we were to use Python for this, we could combine the XML as a structured document?
(see for example: https://github.com/mesosphere/dcos-commons/blob/master/frameworks/hdfs/tests/test_overlay.py#L65)

tools/hive/hadoop-hive/scripts/hive-bootstrap.sh
samvantran: removed `-d`

tools/hive/hadoop-hive/scripts/hadoop-bootstrap.sh
samvantran: I don't think so. This is a script that gets called from `hive-bootstrap.sh` so it'll just continue on afterward

tools/hive/hadoop-hive/scripts/hive-bootstrap.sh
samvantran: Not sure. I'm not very familiar enough w/ Hive.

tools/hive/hadoop-hive/scripts/hive-bootstrap.sh
samvantran: fixed

tools/hive/kerberos/conf/krb5.conf
samvantran: Hm, this is just for testing. Not sure we need to make this configurable.

elezar: My concern is that it is different to how we handle KDC in all our other applications. If we need to move forward with this though, I'm not going to block the PR, but we should consider creating a follow-up ticket to unify this.

tools/hive/kerberos/marathon/hdfs-hive-kerberos.json
```diff
@@ -0,0 +1,39 @@
+{
```
elezar: In other tests we generate these kinds of application definitions on the fly in python -- templating where applicable. Not a blocker, but we could create a follow-up ticket.

tools/hive/kerberos/templates/hdfs-site-kerberos.xml.template
```diff
@@ -0,0 +1,53 @@
...
+    <!-- General HDFS security config -->
+    <property>
+        <name>dfs.block.access.token.enable</name>
+        <value>true</value>
+    </property>
+
+    <!-- NameNode security config -->
+    <property>
+        <name>dfs.namenode.keytab.file</name>
+        <value>/usr/local/hadoop/etc/hadoop/hdfs.keytab</value> <!-- path to the HDFS keytab -->
```
elezar: This is also quite different to how we currently deploy kerberized HDFS. I know there isn't necessary too much overlap, but it would be good to not have to context switch too much when debugging issues with hive / hdfs.

tools/hive/kerberos/templates/hive-site-kerberos.xml.template
```diff
@@ -0,0 +1,48 @@
+<configuration>
+  <!-- Authentication -->
+  <property>
+    <name>hive.server2.authentication</name>
```
elezar: Is `server2` a predefined property of some kind?

PR #290, opened 2018-04-23
[SPARK-717] configuration: enable statsd sink to monitor Spark metrics

hectorj2f:
This PR instruments the service to expose Spark metrics via dcos-metrics.

PR #393, opened 2018-08-17
Update spark-statsd jar to use versions Spark 2.2.1 and Scala 2.11

samvantran:
Resolves [DCOS-40592](https://jira.mesosphere.com/browse/DCOS-40592)

Looks like the statsd jar Hector added previously used older versions of Scala and Spark (2.10 and 1.5 respectively). which resulted in some tests failing w/ `NoSuchMethodErrors`. 

```
spark-env: Skipping metrics configuration: STATSD_UDP_HOST/STATSD_UDP_PORT are not defined
[07:40:53][Step 7/8] Exception in thread "main" java.lang.NoSuchMethodError: scala.Predef$.$conforms()Lscala/Predef$$less$colon$less;scala.Predef$.$conforms()Lscala/Predef$$less$colon$less;
[07:40:53][Step 7/8] 	at org.apache.spark.util.Utils$.getDefaultPropertiesFile(Utils.scala:2097)
[07:40:53][Step 7/8] 	at org.apache.spark.deploy.SparkSubmitArguments$$anonfun$mergeDefaultSparkProperties$1.apply(SparkSubmitArguments.scala:124)
[07:40:53][Step 7/8] 	at org.apache.spark.deploy.SparkSubmitArguments$$anonfun$mergeDefaultSparkProperties$1.apply(SparkSubmitArguments.scala:124)
[07:40:53][Step 7/8] 	at scala.Option.getOrElse(Option.scala:120)
[07:40:53][Step 7/8] 	at org.apache.spark.deploy.SparkSubmitArguments.mergeDefaultSparkProperties(SparkSubmitArguments.scala:124)
[07:40:53][Step 7/8] 	at org.apache.spark.deploy.SparkSubmitArguments.<init>(SparkSubmitArguments.scala:110)
[07:40:53][Step 7/8] 	at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:114)
[07:40:53][Step 7/8] 	at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
[07:40:53][Step 7/8] I0817 07:38:09.268573 32547 executor.cpp:736] Container exited with status 1
```

I rebuilt the jar from https://github.com/vidhyaarvind/spark-statsd to use Spark 2.2.1 and Scala 2.11 and that fixed master branch tests.


susanxhuynh:
Hmm, that's a problem because our Spark package has to support both Docker and UCR containerizers. We have to continue testing the Docker containerizer, at least in some integration tests.

susanxhuynh:
The metrics PR that was merged a while ago: https://github.com/mesosphere/spark-build/pull/290/files
I'm not sure what's causing the problem: is it having the metrics.properties.template file there, or because  of the added jar? The code in spark-env.sh checks if the env vars are defined, so that part seems to working as expected.

elezar:
As @susanxhuynh points out, this changes what we are testing quite dramatically if we are now not exercising the Docker containerizer in our tests. Is the expectation that metrics should work for the Docker containerizer too, or are they mesos only?

samvantran:
@elezar yes metrics [only works with UCR](https://github.com/mesosphere/spark-build/blob/master/conf/spark-env.sh#L86)

Yeah it's strange because not _all_ of the tests failed. Only some do and they seem to [differ across CI jobs](https://teamcity.mesosphere.io/viewLog.html?buildId=1168120&buildTypeId=DcosIo_DcosCommons_Spark_SparkBuildPr&tab=buildResultsDiv)

It seems like leaving it up to the docker containerizer randomly fails. It might be a jar dependency issue. In my experience `NoSuchMethodError` usually happens as a result of jars pulling in different versions so it very well could be the statsD jar pulling in some newer version of Spark and the docker containerizer picks the wrong one?

susanxhuynh:
@samvantran Yeah, probably a jar issue. In Hector's PR (https://github.com/mesosphere/spark-build/pull/290/files), there's a pointer to where the jar comes from: https://github.com/vidhyaarvind/spark-statsd The "build.sbt" file there has Scala version 2.10 which is incompatible with our Spark Scala version (2.11). If it was built with this default, that would explain the problem.

susanxhuynh:
The Spark version could be a problem, too (Spark 1.5 in the "build.sbt").

samvantran:
Okay let me try updating the statsD jar to use Scala 2.11 and Spark 2.2.1

If that doesn't work, we could probably `rm` the jar from the mesos sandbox for non-UCR jobs

samvantran:
Looks like updating the jar did the trick @susanxhuynh @elezar

samvantran:
We have the scale test in two days, should we hold off merging this to master if we think it might negatively affect metrics? @elezar wdyt? 

Another option is to merge and test on a mini-cluster like we normally do before a large scale test.

elezar:
> We have the scale test in two days, should we hold off merging this to master if we think it might negatively affect metrics? @elezar wdyt?

Are we running the tests off master? I thought we add a stub universe?

docker/Dockerfile
```diff
@@ -78,7 +78,8 @@ ADD dist /opt/spark/dist
 ADD krb5.conf.mustache /etc/
 
 # Add external jar for pushing metrics via StatsD (https://github.com/vidhyaarvind/spark-statsd)
-ADD jars/spark-statsd-1.0.0.jar /opt/spark/dist/jars/
+# current version: Spark 2.2.1, Scala 2.11
+ADD jars/spark-statsd-2.2.1-2.11.jar /opt/spark/dist/jars/
```
elezar: Does it make sense to fetch this using the Mesos fetcher instead? (maybe that is out of scope for this PR though).

samvantran:
@elezar it won't affect the stub but the batch marathon apps run off master branch. However they have their own images so we should be fine. I was just being extra cautious.

docker/Dockerfile
samvantran: Yeah, maybe in a different PR but I'm more of the mind 'if it ain't broke don't fix it'

PR #399, opened 2018-08-31
[DCOS-38980] Hive integration test for Sentry CREATETABLE bug fix

susanxhuynh:
Test for running a Hive Spark job against Hive in a (kerberized) Sentry-secured cluster, where a non-admin user has all privileges to a certain database.

The test:
- Sets up a KDC.
- Adds a "alice" principal to the KDC.
- Runs the kerberized Hive Sentry Docker image.
- Grants ALL privileges on Hive database "default" to "alice".
- Prepares the hdfs-site.xml, core-site.xml, and hive-site.xml client config files.
- Runs a Spark job - a Hive job - with "alice" as the Kerberos principal.
-- The job creates several Spark SQL tables with a Hive backend.
-- HDFS is used as an intermediate data store.
-- The job is based on an example from Apache Spark.

This PR also adds support for passing in a client hive-site.xml config file when running a job. The config file may be specified as a Mesos URI in the submit args:
`--conf spark.mesos.uris=http://path/to/hive-site.xml`

Known issues:
- Instead of sleeping for a fixed period, the test should add a readiness check to determine when the Hive image is ready.

### DIFF
--- a/tests/resources/hadoop-hive-kerberos.json
+++ b/tests/resources/hadoop-hive-kerberos.json
@@ -3,6 +3,7 @@
   "cpus": 1,
   "mem": 4096,
   "user": "root",
+  "acceptedResourceRoles": ["slave_public"],
   "container": {
     "type": "MESOS",
     "docker": {


tests/resources/hadoop-hive-kerberos.json
```diff
@@ -0,0 +1,31 @@
+{
+  "instances": 1,
+  "cpus": 1,
+  "mem": 4096,
+  "user": "root",
+  "container": {
+    "type": "MESOS",
+    "docker": {
+      "image": "susanxhuynh/cdh5-hive-kerberos:latest",
```
elezar: Let's add this to the list of docker images that we need to move to the `mesosphere` account.

tests/resources/hdfs-site.xml.template
```diff
@@ -0,0 +1,58 @@
+<configuration>
```
elezar: Can this file not be downloaded from the running hive image?

tests/resources/hive-site.xml.template
```diff
@@ -0,0 +1,26 @@
+<configuration>
```
elezar: Same comment here. Can this not be downloaded from the running hive container?

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+ALICE_USER = "alice"
+ALICE_PRINCIPAL = "{user}@{realm}".format(user=ALICE_USER, realm=sdk_auth.REALM)
+HIVE_APP_ID = "cdh5-hadoop-hive-kerberos"
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+class KerberosSetup:
+    def __init__(self, kerberos_env, hive_agent_ip, hive_agent_hostname):
+        self.kerberos_env = kerberos_env
+        self.hive_agent_ip = hive_agent_ip
```
elezar: Why is the `hive_agent_ip` part of the `KerberosSetup`. Also, should this not be `KerberosConfig`?

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class HadoopSetup:
+    def __init__(self, kerberos_setup, hdfs_base_url, hive_config_url):
+        self.kerberos_setup = kerberos_setup
+        self.hdfs_base_url = hdfs_base_url
+        self.hive_config_url = hive_config_url
+
+
+def _get_agent_ip():
```
elezar: It's not quite clear what this function is supposed to do? Get the IP of the first private agent?

What about using this: https://github.com/mesosphere/dcos-commons/blob/master/testing/sdk_agents.py#L32

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+@pytest.fixture(scope='module')
+def kerberos_setup():
+    try:
+        hive_agent_ip = _get_agent_ip()
+
+        ok, stdout = sdk_cmd.agent_ssh(hive_agent_ip, "echo $HOSTNAME")
+        if not ok:
+            raise Exception("Failed to get agent hostname, stdout: {}".format(stdout))
+        log.info("got hostname: '{}'".format(stdout))
+        hive_agent_hostname = stdout.rstrip()
```
elezar: I don't follow how this is the hostname of the agent that is actually running the hive image, or am I missing something?

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+@pytest.fixture(scope='module')
+def hadoop_setup(kerberos_setup):
+    try:
+        # TODO: replace with Marathon readiness check in kdc app?
+        #log.info("Waiting for kdc.marathon.mesos to resolve ...")
+        #time.sleep(10)
+
+        # Run the Hive Docker image in Marathon
+        curr_dir = os.path.dirname(os.path.realpath(__file__))
+        app_def_path = "{}/resources/hadoop-hive-kerberos.json".format(curr_dir)
```
elezar: For many of our other marathon applications, we specify the app config directly in the python code. Is there a motivation for doing it like this?

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+        with open(app_def_path) as f:
+            hive_app_def = json.load(f)
+        hive_app_def["id"] = HIVE_APP_ID
+        hive_app_def["constraints"] = [["hostname", "IS", kerberos_setup.hive_agent_ip]]
+        sdk_marathon.install_app(hive_app_def)
+        # Wait for Sentry and Hive to be ready
+        # TODO: replace with Marathon health check
```
elezar: Yes. Please do add *at least* a wait for deployment.

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+        #log.info("Waiting for kdc.marathon.mesos to resolve ...")
+        #time.sleep(10)
+
+        # Run the Hive Docker image in Marathon
+        curr_dir = os.path.dirname(os.path.realpath(__file__))
+        app_def_path = "{}/resources/hadoop-hive-kerberos.json".format(curr_dir)
+        with open(app_def_path) as f:
+            hive_app_def = json.load(f)
+        hive_app_def["id"] = HIVE_APP_ID
+        hive_app_def["constraints"] = [["hostname", "IS", kerberos_setup.hive_agent_ip]]
```
elezar: What if the host does not have the required resources? Since the requirements are pretty low it may not be too much of an issue though.

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+        sdk_marathon.install_app(hive_app_def)
+        # Wait for Sentry and Hive to be ready
+        # TODO: replace with Marathon health check
+        log.info("Sleeping 120s ...")
+        time.sleep(120)
+
+        # HDFS client configuration
+        core_site_url = _upload_hadoop_config("core-site.xml", kerberos_setup.hive_agent_hostname)
```
elezar: IIRC these are already part of the Docker image?

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+        _upload_hadoop_config("hdfs-site.xml", kerberos_setup.hive_agent_hostname)
+        hive_config_url = _upload_hadoop_config("hive-site.xml", kerberos_setup.hive_agent_hostname)
+        hdfs_base_url = os.path.dirname(core_site_url)
```
elezar: It seems like we're mixing paths and URLS here?

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def _grant_hive_privileges():
+    rc, _, err = sdk_cmd.marathon_task_exec(HIVE_APP_ID, "/etc/grant-hive-privileges.sh")
+    if rc != 0:
+        raise Exception("Hive script failed with code {}: {}".format(rc, err))
+
+
+@pytest.mark.xfail(sdk_utils.is_strict_mode(),
+    reason="hadoop_setup requires root access; won't work in strict mode")
```
elezar: I don't quite think that this reasoning is valid? Which part of the application requires `root` access? Can we not then run the application as the `root` user (giving Marathon the required permissions to launch it successfully).

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+@pytest.mark.sanity
+def test_hive(hadoop_setup, setup_spark):
+    # Job writes to this hdfs directory
+    _hdfs_cmd("chmod a+w /")
+
+    _grant_hive_privileges()
+
+    kerberos_args = [
+        "--kerberos-principal", ALICE_PRINCIPAL,
+        "--keytab-secret-path", "/__dcos_base64___keytab",
```
elezar: Was there some issue with regards to accessing secrets at the root level?

tests/test_hive.py
```diff
@@ -0,0 +1,190 @@
...
+        "--conf", "spark.mesos.driverEnv.SPARK_USER={}".format(spark_utils.SPARK_USER)]
+
+    submit_args = [
+        "--class", "HiveFull",
+        "--conf", "spark.mesos.uris={}".format(hadoop_setup.hive_config_url),
+        ] + kerberos_args
+
+    spark_utils.run_tests(app_url=spark_utils.dcos_test_jar_url(),
+                          app_args="thrift://{}:9083".format(hadoop_setup.kerberos_setup.hive_agent_hostname),
```
elezar: Should we not read this from the `hive-site.xml` file instead?

PR #402, opened 2018-08-31
[SPARK-717] configuration: enable statsd sink to monitor Spark metric…

samvantran:
Backport of #290 && #393 

This PR instruments the service to expose Spark metrics via dcos-metrics.

### DIFF
--- a/conf/metrics.properties.template
+++ b/conf/metrics.properties.template
@@ -1,0 +1,9 @@
+
+# Enable StatsdSink for all instances by class name
+*.sink.statsd.class=org.apache.spark.metrics.sink.StatsDSink
+*.sink.statsd.prefix=spark
+*.sink.statsd.host=<STATSD_UDP_HOST>
+*.sink.statsd.port=<STATSD_UDP_PORT>
+
+# Enable JVM metrics source for all instances (master, worker, driver and executor) by class name
+*.source.jvm.class=org.apache.spark.metrics.source.JvmSource
--- a/conf/spark-env.sh
+++ b/conf/spark-env.sh
@@ -12,7 +12,7 @@ mkdir -p "${HADOOP_CONF_DIR}"
 [ -f "${MESOS_SANDBOX}/hdfs-site.xml" ] && cp "${MESOS_SANDBOX}/hdfs-site.xml" "${HADOOP_CONF_DIR}"
 [ -f "${MESOS_SANDBOX}/core-site.xml" ] && cp "${MESOS_SANDBOX}/core-site.xml" "${HADOOP_CONF_DIR}"
 
-
+SPARK_WORKING_DIR=${PWD}
 
 cd $MESOS_SANDBOX
 
@@ -85,6 +85,14 @@ if [ -n "${STATSD_UDP_HOST}" ] && [ -n "${STATSD_UDP_PORT}" ]; then
 else
     echo "spark-env: Skipping metrics configuration: STATSD_UDP_HOST/STATSD_UDP_PORT are not defined" >&2
     echo "spark-env: StatsD metrics require Mesos UCR. For dispatcher metrics, enable the 'UCR_containerizer' option. For driver metrics, include '--conf spark.mesos.containerizer=mesos' in your run"
+fi
+
+if [ -n "${STATSD_UDP_HOST}" ] && [ -n "${STATSD_UDP_PORT}" ]; then
+    sed -e "s/<STATSD_UDP_HOST>/${STATSD_UDP_HOST}/g" \
+        -e "s/<STATSD_UDP_PORT>/${STATSD_UDP_PORT}/g" \
+        ${SPARK_WORKING_DIR}/conf/metrics.properties.template >${SPARK_WORKING_DIR}/conf/metrics.properties
+else
+    echo "spark-env: No STATSD_UDP environment variables were defined" >&2
 fi
 
 # Options read when launching programs locally with
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -78,7 +78,8 @@ ADD dist /opt/spark/dist
 ADD krb5.conf.mustache /etc/
 
 # Add external jar for pushing metrics via StatsD (https://github.com/vidhyaarvind/spark-statsd)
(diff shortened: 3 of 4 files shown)


conf/spark-env.sh
```diff
@@ -87,6 +87,14 @@ else
     echo "spark-env: StatsD metrics require Mesos UCR. For dispatcher metrics, enable the 'UCR_containerizer' option. For driver metrics, include '--conf spark.mesos.containerizer=mesos' in your run"
 fi
 
+if [ -n "${STATSD_UDP_HOST}" ] && [ -n "${STATSD_UDP_PORT}" ]; then
+    sed -e "s/<STATSD_UDP_HOST>/${STATSD_UDP_HOST}/g" \
```
elezar: I know this is a backport PR, but would it make sense to at some point use mustache templating instead of non-standard `<VAR>` tags?

conf/spark-env.sh
```diff
@@ -12,7 +12,7 @@ mkdir -p "${HADOOP_CONF_DIR}"
 [ -f "${MESOS_SANDBOX}/hdfs-site.xml" ] && cp "${MESOS_SANDBOX}/hdfs-site.xml" "${HADOOP_CONF_DIR}"
 [ -f "${MESOS_SANDBOX}/core-site.xml" ] && cp "${MESOS_SANDBOX}/core-site.xml" "${HADOOP_CONF_DIR}"
 
-
+SPARK_WORKING_DIR=${PWD}
```
elezar: Is the CWD explicitly set? I know that in TensorFlow I saw some undefined behaviour if the `WORKDIR` was not set in a docker image.

conf/metrics.properties.template
```diff
@@ -0,0 +1,9 @@
+
+# Enable StatsdSink for all instances by class name
+*.sink.statsd.class=org.apache.spark.metrics.sink.StatsDSink
+*.sink.statsd.prefix=spark
+*.sink.statsd.host=<STATSD_UDP_HOST>
```
elezar: See comment regarding templating in this file.

docker/Dockerfile
```diff
@@ -77,6 +77,9 @@ RUN ln -s /var/lib/runit/service/nginx /etc/service/nginx
 ADD dist /opt/spark/dist
 ADD krb5.conf.mustache /etc/
 
+# Add external jar for pushing metrics via StatsD (https://github.com/vidhyaarvind/spark-statsd)
+ADD jars/spark-statsd-1.0.0.jar /opt/spark/dist/jars/
```
elezar: Would it make sense to include this from `downloads.mesosphere` instead, or is there precedent  to add it like this. What worries me is that there is no indication as to how the jar was built.

docker/runit/service/spark/run
```diff
@@ -31,6 +31,14 @@ function set_log_level() {
         ${SPARK_DIST_PATH}/conf/log4j.properties.template >${SPARK_DIST_PATH}/conf/log4j.properties
 }
 
+function set_statsd_address() {
```
elezar: NIT: This is a duplication of the templating in `spark-env.h`.

conf/spark-env.sh
elezar: For that matter, don't we have `bootstrap` in this image to do the templating?

conf/spark-env.sh
samvantran: I don't think its explicitly set but it looks like [this spark script](https://github.com/mesosphere/spark/blob/custom-branch-2.2.1-X/bin/load-spark-env.sh#L20) finds this file and loads all its envvars for every spark job so the PWD would at least point inside Spark?

docker/Dockerfile
samvantran: Not sure about precedence but I do think this is easier (since we don't have an equivalent SDK `svc.yml` file). However the link in the comment does point to the project with instructions to build the jar.

If we went the `downloads.mesosphere` route: 
- we would need to add a file (or files) to build the jar
- add a jenkins job to build and upload the jar to the mesosphere acct 
- update Spark's marathon cmd to mv the file to directory (`opt/spark/dist/jars`)
- (not sure if there are more steps beyond that)

Not saying it's hard but it would add more work for the same result. Not sure its worth the effort?

docker/runit/service/spark/run
samvantran: Good catch. We could just delete?

samvantran:
@elezar I've created https://jira.mesosphere.com/browse/DCOS-42223 to track/address your comments in this PR. Since this is a backport, can we merge and address in RI-6 instead?

conf/spark-env.sh
elezar: We use `bootstrap` to apply templating to config files in other data services. 

Look at https://github.com/mesosphere/dcos-commons/blob/master/sdk/bootstrap/main.go#L338 for example. Note that we may want to add a filename parameter to the utility to allow for local usage instead of mapping this through the envvars.

docker/runit/service/spark/run
elezar: I don't know if we can just delete it. Are there cases where this code is executed and not `spark-env.h`?

conf/spark-env.sh
elezar: I don't see how that script changes the current working directory though? I was mainly asking because this seems to introduce an envvar that is not related to the metrics and is not used elsewhere.

Dropping this Envvar and using relative paths in the templating below would have the same effect.

samvantran:
There aren't automated tests in this PR but we did test this heavily with the mixed workload testing. The perm stub that we used for each run included this fix so that we could get spark metrics

samvantran:
@elezar, just a quick FYI, Mahendra, Vishnu and I had a quick meeting to discuss this PR. Adding automated tests are important and will be done in the next RI. However, we'll merge this as is and declare it as a beta feature because 
- it is not enabled by default (users would have to enable UCR to get metrics) and
-  it has already been tested many times in the MWTs.

As mentioned above, we'll use ticket [DCOS-42223](https://jira.mesosphere.com/browse/DCOS-42223) to track the fixes required

PR #405, opened 2018-09-05
[DCOS-41319] Use standard mesosphere manifest URI

samvantran:
Fixes: [DCOS-41319](https://jira.mesosphere.com/browse/DCOS-41319)

This updates the `manifest.json` URI to the standard `downloads.mesosphere.com/spark/assets` path.

## How was this patch tested?
- [This TC build](https://teamcity.mesosphere.io/viewLog.html?buildId=1195187&tab=buildResultsDiv&buildTypeId=DcosIo_DcosCommons_Spark_SparkBuildPr) passed successfully


manifest.json
```diff
@@ -2,16 +2,16 @@
     "spark_version": "2.2.1",
     "default_spark_dist": {
         "hadoop_version": "2.6",
-        "uri": "svt-dev.s3.amazonaws.com/spark/spark-2.2.1-bin-2.6.5-DCOS-38138.tgz"
+        "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.2.1-3-SNAPSHOT-bin-2.6.tgz"
```
elezar: This is still a `SNAPSHOT` version. Is that intentional?

samvantran: Yes I still have to merge in Susan's Hive changes before we release the spark  package. Once that's in I'll create the `2.2.1-3` tarball.

elezar: I suppose my question is whether this is generated by the build process -- so that we don't have to manually update this each time. I'm not saying that we have to do it now, but it may make sense to file a ticket to get this working in a better way.

samvantran: That's a good point. Created ticket for that future work [DCOS-41439](https://jira.mesosphere.com/browse/DCOS-41439)

PR #408, opened 2018-09-19
Branch 3.0

soumasish:
Beefed up multi_args_test to check for spark app name.


soumasish:
Makes sense, I'll delete this PR and create a new one.

PR #410, opened 2018-09-19
[DCOS-42237] Add manifest-dist back to Makefile (#409)

samvantran:
Backport of PR #409 

Adds `manifest-dist` back to Makefile to fix the jenkins-package-publish script


samvantran:
The short-term solution is that we'll (ideally) be releasing monthly so we'll be able to catch these things sooner than later should a breaking change occur.

But eventually, I think we'll want to follow dcos-commons entirely. Maybe we should have a monthly sync with their test/helper files so that we never stray too far.

PR #411, opened 2018-09-21
[COPS-3550] Replace hardcoded VIP with service.name

samvantran:
Resolves [COPS-3550](https://jira.mesosphere.com/browse/COPS-3550) && [DCOS-39751](https://jira.mesosphere.com/browse/DCOS-39751)

Prior to this fix, VIP endpoints were hardcoded to `spark-dispatcher` and if you had multiple spark dispatchers running you would not be able to reach any specific one.

## How was this patch tested?
- Added an integration test for verifying that unique VIPs are reachable

### To test manually:
- upload a stub to your test cluster (you can the latest stub created from this PR's TC job)
- install one or more spark dispatchers
- verify that the VIPs in the endpoints tab are unique (and not set to `spark-dispatcher`)

see screenshots:

### Before
screen shot 2018-09-21 at 1 58 57 pm

### After
screen shot 2018-09-21 at 1 59 11 pm


universe/marathon.json.mustache
```diff
@@ -60,23 +60,23 @@
             "protocol": "tcp",
             "name": "dispatcher",
             "labels": {
-                "VIP_0": "spark-dispatcher:7077"
+                "VIP_0": "{{service.name}}:7077"
```
elezar: This is a breaking change for existing customers. This should be `{{service.name}}-dispatcher`. cc @vishnu2kmohan

samvantran: Good catch. We should maintain this for our 2.2.1-x packages but do we want to keep this when we bump base-tech to spark 2.3?

samvantran: Actually, this would only maintain behavior for clients who installed spark packages as `spark`. If they named it anything else (eg foldered names), this will still break once they update the package version.

PR #413, opened 2018-09-24
Added test for spark name

soumasish:


soumasish:
##What changes were proposed in this pull request?
Shell escaped the name passed to spark-submit and change how conf attributes are shell escaped.

##How was this patch tested?
This feature has been tested manually running the spark distro on DC/OS with the use case described in the issue with the sparkPi application with a custom name.

PR #419, opened 2018-10-04
[DCOS-43232] Add infrastructure cleanup option to run.sh

porridge:
Also put a timestamp in the log name to make it safe to re-run the
script with the same args.


porridge:
Note: we should update the runbook doc once this is merged.

scale-tests/run.sh
```diff
@@ -70,7 +70,7 @@ readonly CONTAINER_NAME="${TEST_NAME}"
 readonly CONTAINER_SSH_AGENT_EXPORTS=/tmp/ssh-agent-exports
 readonly CONTAINER_SSH_KEY=/ssh/key
 readonly IMAGE_NAME="mesosphere/dcos-commons:${TEST_NAME}"
-readonly LOG_FILE="${TEST_NAME}.log"
+readonly LOG_FILE="${TEST_NAME}_$(date +%Y%m%d-%H%M%S).log"
```
mpereira: Since the log file is always appended to, and log lines have timestamps I think it's better to only have one log file per test.

mpereira:
Thanks @porridge, this is going to be handy! I just left a comment about the log file name change.

scale-tests/run.sh
porridge: Hm, I did this because it was not obvious whether keeping the same filename would clobber the contents, so I thought it is not safe to reuse it. I'm afraid others might have the same worry in the future...

scale-tests/configs/2018-10-04-mwst8.env
```diff
@@ -11,6 +11,7 @@ SECURITY="strict"
 # Test configuration ###########################################################
 
 SHOULD_INSTALL_INFRASTRUCTURE=true
+SHOULD_UNINSTALL_INFRASTRUCTURE=false
```
elezar: Is there some interaction between `SHOULD_INSTALL_INFRASTRUCTURE` and `SHOULD_UNINSTALL_INFRASTRUCTURE`?

porridge: No, just similar name. Do you want it moved to the bottom?

scale-tests/run.sh
```diff
@@ -448,6 +450,18 @@ else
   log 'Skipping running of GPU batch jobs'
 fi
 
+if [ "${SHOULD_UNINSTALL_INFRASTRUCTURE}" = true ]; then
+  log 'Uninstalling infrastructure'
+  start_time=$(date +%s)
```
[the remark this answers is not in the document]
porridge: I don't think we need them here, we're doing arithmetic on the result.

scale-tests/run.sh
porridge: So WDYT @mpereira and @elezar - log per execution or per test?

mpereira: I think it's better to have one per *test*. Maybe we can leave a comment mentioning that the file is only appended to and not truncated.

scale-tests/run.sh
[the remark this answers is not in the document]
mpereira: Yup.

scale-tests/run.sh
porridge: I cannot think of a good place to put the comment. When I made this change it was because after having ran the script once, and before running it the second time, I was wondering what test name to use, looked at the output from the first run, and there one of the last things was uploading of the log to s3. I immediately thought "oops, I better not reuse the same test name because I will clobber the log on s3".

scale-tests/configs/2018-10-04-mwst8.env
porridge: Right, if you select any other steps, then yes. Any ideas how to make this clearer? `SHOULD_UNINSTALL_INFRASTRUCTURE_AT_THE_END`?

scale-tests/run.sh
mpereira: > I don't think output mentioning append is too useful as there is a LOT of output on screen already.

I meant writing a comment in the code.

Now that I think again, I think prepending the date to the log file names is a good move. If two people run the script, unless the second person downloaded a possibly existing log in S3 for the same test before running the script, as the script run finishes it will overwrite that existing S3 log file created when the first person ran the script. We could make the script download the log file from S3 as a first step, but I think having multiple log files is better.

We could also append the user name so that the logs in a test's S3 bucket look like:
```
20181004T120000Z_marcin.log
20181004T133000Z_marcin.log
20181004T153000Z_karan.log
20181004T163000Z_sam.log
``` 

Let's use UTC and a ISO8601-ish format?
```bash
readonly TEST_NAME="$(date -u +"%Y%m%dT%H%M%SZ")_$(whoami).log"
```

If you prefer we can also keep this change outside the scope of this PR and do it in a different one, @porridge.

scale-tests/configs/2018-10-04-mwst8.env
elezar: I was also thinking of renaming the envvar. As I mentioned, not critical though.

elezar:
Seems as if the timestamp question has been resolved. We can always add the user in a different pr.

porridge:
Comments addressed, I also set the timezone to UTC for sanity.
Merging.

PR #424, opened 2018-10-12
Update publish_aws to name assets similar to dcos-commons

samvantran:
## What changes are proposed in this PR? 
Resolves [DCOS-42789](https://jira.mesosphere.com/browse/DCOS-42789)

This PR updates the `publish_aws.py` script to name permanent stubs the same way dcos-commons does. The code change was taken from the identically-named [dcos-commons script](https://github.com/mesosphere/dcos-commons/blob/master/tools/publish_aws.py#L130)

This script now looks for an envvar `S3_DIR_NAME` which is set in the jenkins build configuration [here](https://jenkins.mesosphere.com/service/jenkins/view/Infinity/job/spark/job/spark-build/job/spark-build-release/configure) as `S3_DIR_NAME=assets`.

previous stubs were named:
`https://infinity-artifacts.../permanent/spark/20180303-002532-DvZbZFyeM14X1qgF/...`
now they're named:
 `https://infinity-artifacts.../permanent/spark/assets/<version>/...`


tools/publish_aws.py
```diff
@@ -143,6 +120,38 @@ def print_help(argv):
...
 
 
+def s3_urls_from_env(package_name):
+    s3_bucket = os.environ.get("S3_BUCKET") or "infinity-artifacts"
+    logger.info("Using artifact bucket: {}".format(s3_bucket))
+
+    s3_dir_path = os.environ.get("S3_DIR_PATH") or "autodelete7d"
+    s3_dir_name = os.environ.get("S3_DIR_NAME")
+    if not s3_dir_name:
+        s3_dir_name = "{}-{}".format(
```
soumasish: Not specific to this PR, but strftime will only parse a string into time when the string is in the specific format. While this might always be the case here, it's a bit limiting. dateutil.parse will convert any string to a python datetime object.

tools/publish_aws.py
```diff
@@ -143,6 +120,38 @@ def print_help(argv):
...
+            time.strftime("%Y%m%d-%H%M%S"),
```
soumasish: Not specific to this PR, but strftime will only parse a string into time when the string is in the specific format. While this might always be the case here, it's a bit limiting. dateutil.parse will convert any string to a python datetime object.

elezar: Note: `strftime` doesn't parse a string it generates a string from the current time. `strptime` parses the time.

samvantran:
Ended up copying the entire from `dcos-commons` @elezar

PR #426, opened 2018-10-15
Add test parameters.

mpereira:
These are the config files for both the dry-run and the actual MWT #9 test.


elezar:
> We're getting to a point where many of these settings are essentially copied over minus one or few changes. Might be worth creating a default template that gets overwritten with specified envvars? Similar to how we deploy services across soak clusters (or even more generally, how we deploy SDK FWs). WDYT? @mpereira @elezar

I would definitely argue for a bit more consistency between soak and the MWT. I will try and document this before the next one.

mpereira:
> @elezar: It is my understanding that this worked as expected.

That is correct.

mpereira:
> @samvantran: We're getting to a point where many of these settings are essentially copied over minus one or few changes. Might be worth creating a default template that gets overwritten with specified envvars? Similar to how we deploy services across soak clusters (or even more generally, how we deploy SDK FWs). WDYT?

I think some of the parameters don't need to be in each test parameter files, like the output filenames, so we could move those out as default parameters.

In one hand it is nice to have all parameters for a workload test defined in a single place. In addition to not have to deduce a test's parameters based on multiple files (involving script logic, conditionals etc.) it presents a straightforward history for test runs' parameters.

On the other hand this is inconsistent with how things are handled in Soak clusters, which is the direction we'd like to move these tests towards. When the time comes I think it will be valuable to stop and think carefully about how we can optimally define this workflow.

PR #429, opened 2018-10-17
[DCOS-42550] Helper script for creating stub universe from local Spark sources

akirillov:
This PR introduces a small helper script which allows creating a stub universe for Spark from developer's machine and doesn't require any extra-steps like uploading Spark distribution to S3 or running TeamCity job.


README.md
```diff
@@ -55,6 +55,33 @@ make stub-universe-url
...
+Options:
+--spark-dist-dir <absolute path>  Mandatory. Absolute path to Spark project sources used to build and/or upload Spark archive
+--hadoop-version <version>        Optional. Hadoop version to build Spark with. Default: 2.7
+--docker-dist-image <image tag>   Optional. Target Docker image to publish. Default: mesosphere/spark-dev:<git commit sha>
+--rebuild-docker                  If Docker image should be rebuilt. Default: false
+--rebuild-spark                   If Spark distribution should be rebuilt. Default: false
+--help                            Print this message
+```
+
+Assuming that Spark sources are located at `/usr/projects/mesosphere/spark` script invocation will look like that:
```
samvantran: Small grammar nit. Can you change to:
```
Assuming Spark source code is located at `/usr/projects/mesosphere/spark` script invocation will look like:
```

akirillov:
@elezar if you look into the script it just moves spark dist to a location used by `Makefile` and then calls specific targets. I think that initially Spark sources and integration tests/builds lived in the same repo, that's why `Makefile` makes some assumptions about the directory structure and even downloads Spark sources if `spark` folder is missing. The bottom line is that this solution is temporary and we should definitely use a unified approach from `dcos-commons` when implementing Spark service with SDK.

README.md
akirillov: Thanks, fixed

PR #430, opened 2018-10-18
Fixes to publish_local_spark.sh to make it compatible with other bash versions

akirillov:
This fix makes iteration over the script arguments stable across different shells. The issue was related to the `shift` inside an array iterator which led to the incorrect assignment of the script variables.

To verify the change:
```
# from the root of this project
./publish_local_spark.sh --spark-dist-dir <path to mesosophere spark> --docker-dist-image user/spark-dev:test --rebuild-spark --rebuild-docker --hadoop-version 2.7

#expect the following lines in the beginning of the output:

Running with parameters:
SPARK_DIST_DIR=<path to mesosophere spark>
HADOOP_VERSION=2.7
DOCKER_DIST_IMAGE=user/spark-dev:test
REBUILD_DOCKER=true
REBUILD_SPARK=true
PRINT_HELP=false
```


akirillov:
@soumasish please verify, that the updated script runs on your machine without issues

samvantran:
The script as is does not work inside an ubuntu container. Couple things:
- `make` cmd is not available
- `docker` cmd is not available
- docker login is required
- maws li is required 
- starting container with path to spark repo is required

there is the makefile target `make docker-build` which builds a stable docker image we can use. For me, the steps to make this work required

```bash
$ <login with maws>
$ make docker-build
$ docker run --rm -it -v $(pwd):/spark-build -v $(pwd)/../spark:/spark -v ${HOME}/.aws/credentials:/root/.aws/credentials:ro -v /var/run/docker.sock:/var/run/docker.sock -v /usr/local/bin/docker:/usr/bin/docker $(cat docker-build) bash`
$ docker login
$ cd spark-build/
$ ./publish_local_spark.sh --spark-dist-dir /spark --docker-dist-image <username>/spark-dev:test --rebuild-spark --rebuild-docker --hadoop-version 2.7
```
It works but it is quite a manual process. Can we streamline this a little more? Maybe we can build docker image outside of the container and pass that as an envvar or issue `container exec` commands.

publish_local_spark.sh
```diff
@@ -101,4 +113,8 @@ mkdir -p build/dist
 cp ${SPARK_DIST} build/dist
 
 echo "Building stub universe and Uploading Spark Distribution to it"
-DOCKER_DIST_IMAGE=${DOCKER_DIST_IMAGE} make stub-universe-url
\ No newline at end of file
+if [ -n "${DOCKER_DIST_IMAGE}" ]; then
+    export DOCKER_DIST_IMAGE=${DOCKER_DIST_IMAGE}
+fi
+
+make stub-universe-url
```
elezar: Nit: Newline at end of file.

akirillov:
@samvantran the container example was aiming to demonstrate that script's parameters are parsed properly in Ubuntu Docker container to verify that it (parameters parsing) works on Linux. The script wasn't supposed to complete successfully, neither docker container is needed for running it in the future.

PR #431, opened 2018-10-23
[DCOS-39730] Make Spark dispatcher task output Spark process stdout.

mpereira:
Spark schedulers are started by the init.sh script in a Docker container run by Mesos as a task. [Mesos captures tasks' stdout and writes them to "stdout" files in the task sandbox](http://mesos.apache.org/documentation/latest/logging/#containers). Since runit's `runsvdir` doesn't output a process's output to standard out, and instead has [its own logging setup](http://smarden.org/runit/svlogd.8.html) Mesos didn't have anything to capture so it couldn't create "stdout" files.

This PR makes the runit process tail the Spark process output captured by `svlogd` so that Mesos can capture it and write "stdout" files.

I tested this manually to make sure it works, but still want to push an integration test for it.

Spark with stdout:
spark-with-stdout
 
Making sure stdout follows the runit log files even when they rotate:
spark-with-stdout-trace

TODO:
- [ ] Integration test


docker/runit/init.sh
```diff
@@ -32,5 +32,19 @@ sed -i "s,<PROTOCOL>,${SPARK_SSL_PROTOCOL}," /etc/nginx/conf.d/spark.conf
...
+# Start runit processes (spark and nginx).
+exec runsvdir -P /etc/service &
+
+# The Spark process stdout is logged by runit's svlogd to a file named
+# "current".
+readonly SPARK_STDOUT_FILE="${MESOS_SANDBOX}"/spark/current
+
+# Wait for Spark process log file to exist.
+while [ ! -f "${SPARK_STDOUT_FILE}" ]; do
+  sleep 1
```
akirillov: won't this cause an infinite loop in case dispatcher fails to start and doesn't create a file?

docker/runit/init.sh
```diff
@@ -32,5 +32,19 @@ sed -i "s,<PROTOCOL>,${SPARK_SSL_PROTOCOL}," /etc/nginx/conf.d/spark.conf
...
+done
+
+# Tail Spark process log file handled by runit. We do this because the output
+# from this script (which is run in the Marathon app's Docker container) is
+# logged by Mesos to the "stdout" files in the task sandbox.
+tail -F "${SPARK_STDOUT_FILE}"
```
akirillov: It looks like the main container process is now `tail` while service is running in the background. In case of a service failure `tail` will still be running. Wouldn't it be better to either run `tail` in the background or create a symlink for `$SPARK_STDOUT_FILE` like that: `ln -sf /dev/stdout $SPARK_STDOUT_FILE`

samvantran: +1. Either approach works and will allow you to get rid of the sleep (and potential infinite loop)

mpereira: I think running tail in the background works and is a better approach. The symlink approach doesn't work because the runit log files are rotated:
```
+ echo 'line 1'
+ cat current
line 1
+ ln -sf current stdout
+ cat current
line 1
+ cat stdout
line 1
+ echo 'line 2'
+ cat current
line 1
line 2
+ cat stdout
line 1
line 2
+ mv current current.1
+ echo 'line 3'
+ cat current
line 3
+ cat stdout
line 3
+ rm -rf current current.1
+ unlink stdout
```

mpereira: https://github.com/mesosphere/spark-build/pull/431/commits/9d0550a440d835e5ffce8f606639b052ba56251c

samvantran:
Can you also create a backport PR for `branch-2.4`?

mpereira:
@samvantran will do, also for `branch-2.3`.

PR #433, opened 2018-10-24
Spark History Server healthcheck update from MESOS_HTTP to COMMAND

akirillov:
**Change description**
`MESSOS_HTTP` health check stopped working with Spark 2.3 History Server which binds to a private ip instead of `0.0.0.0` or `127.0.0.1`. This resulted in Spark History Server healthcheck failures in Marathon which prevented it from starting.

Excerpt from the documentation, section [Current Limitations](https://mesos.readthedocs.io/en/latest/health-checks/):
> HTTP(S) and TCP health checks use 127.0.0.1 as target IP. As a result, if tasks want to support HTTP or TCP health checks, they should listen on the loopback interface in addition to whatever interface they require (see [MESOS-6517](https://issues.apache.org/jira/browse/MESOS-6517)).

This patch switches from using of `MESOS_HTTP` to `COMMAND` health check which greps server logs for errors.

**Related Apache Mesos issue**
[[MESOS-6517] Health checking only on 127.0.0.1 is limiting](https://issues.apache.org/jira/browse/MESOS-6517)


history/universe/marathon.json.mustache
```diff
@@ -56,9 +56,8 @@
 {{/security.kerberos.keytab}}
     "healthChecks": [
         {
-            "portIndex": 0,
-            "protocol": "MESOS_HTTP",
-            "path": "/",
+            "protocol": "COMMAND",
```
elezar: What about setting the `SPARK_LOCAL_IP` to `0.0.0.0` when launching the history server?

elezar: See https://jaceklaskowski.gitbooks.io/mastering-apache-spark/spark-history-server.html

akirillov: It looks like `0.0.0.0` is the default one but users might want to provide a specific binding (to private ip specifically). In case `SPARK_LOCAL_IP` is different from `0.0.0.0` or `127.0.0.1` `MESOS_HTTP` check will not work. Also, if a node has multiple network interfaces it doesn't look great to bind to all of them from the security perspective.

PR #437, opened 2018-10-25
Spark Dispatcher and History Server health check replacement for MESOS_HTTP

akirillov:
**Change description**
This patch is a continuation of #433 and introduces a behaviour similar to `MESOS_HTTP` check but for a custom ip address.

**Related Apache Mesos issue**
[[MESOS-6517] Health checking only on 127.0.0.1 is limiting](https://issues.apache.org/jira/browse/MESOS-6517)


tools/airgap_linter.py
```diff
@@ -53,6 +53,7 @@ def is_bad_uri(uri, file_name):
         ".mesos:",
         ".mesos/",
         "$MESOS_CONTAINER_IP",
+        "$LIBPROCESS_IP",
```
elezar: I know that this is not ideal, but we should consider the other forms of environment variable usage too:
* `${LIBPROCESS_IP}`
* `{{LIBPROCESS_IP}}` (for mustache templating)

We will also need to make this change in the `dcos-commons` repo to ensure that this is not overwritten when we update the tooling.

akirillov:
@elezar I added different forms of `LIBPROCESS_IP` spelling into the linter + added similar health check to Spark 
Dispatcher to keep everything aligned.

docker/Dockerfile
```diff
@@ -18,7 +18,7 @@
 # docker build -t spark:git-`git rev-parse --short HEAD` .
 
 # Basing from Mesos image so the Mesos native library is present.
-FROM ubuntu:14.04
+FROM ubuntu:16.04
```
porridge: Why not `18.04`?

docker/Dockerfile
```diff
@@ -18,7 +18,7 @@
 # docker build -t spark:git-`git rev-parse --short HEAD` .
 
 # Basing from Mesos image so the Mesos native library is present.
```
porridge: This comment is bitrotten, please remove it.

history/universe/marathon.json.mustache
```diff
@@ -57,7 +57,7 @@
     "healthChecks": [
         {
             "protocol": "COMMAND",
-            "command": { "value": "if [ -n \"$(tail -n500 $MESOS_SANDBOX/stderr | grep ERROR )\" ]; then echo 'Spark History Server logs contain errors'; exit 1; else exit 0; fi" },
+            "command": { "value": "echo \"healthcheck: running curl -s --fail http://$LIBPROCESS_IP:$PORT0/api/v1/version\ > /dev/null"; curl -s --fail \"$LIBPROCESS_IP:$PORT0/api/v1/version\" > /dev/null" },
```
porridge: The quoting here looks wrong:
```
echo \"healthcheck: running curl -s --fail http://$LIBPROCESS_IP:$PORT0/api/v1/version\ > /dev/null"
```
If you escape the leading quote, then the trailing one should be escaped too. Yet you escape a space.
Also echoing this into `/dev/null` isn't very useful :-)
Also, `-o /dev/null` is slightly less awkward in cases like this than the redirection.

tools/airgap_linter.py
```diff
@@ -54,6 +54,9 @@ def is_bad_uri(uri, file_name):
         ".mesos/",
         "$MESOS_CONTAINER_IP",
         "${MESOS_CONTAINER_IP}",
+        "$LIBPROCESS_IP",
```
porridge: Has this been ported upstream?

universe/marathon.json.mustache
```diff
@@ -121,9 +121,8 @@
 
     "healthChecks": [
         {
-            "portIndex": 2,
-            "protocol": "MESOS_HTTP",
-            "path": "/",
+            "protocol": "COMMAND",
```
porridge: Same comments as above.

akirillov:
@porridge PR is cleaned up, all comments addressed. a new PR related to `airgap_linter` changes is created in dcsos-commons https://github.com/mesosphere/dcos-commons/pull/2737

history/universe/marathon.json.mustache
```diff
@@ -57,7 +57,7 @@
     "healthChecks": [
         {
             "protocol": "COMMAND",
-            "command": { "value": "if [ -n \"$(tail -n500 $MESOS_SANDBOX/stderr | grep ERROR )\" ]; then echo 'Spark History Server logs contain errors'; exit 1; else exit 0; fi" },
+            "command": { "value": "curl --fail -s -o /dev/null  \"http://$LIBPROCESS_IP:$PORT0/api/v1/version\"" },
```
porridge: I'd drop the inside quotes, I don't think we need them, since neither env var is allowed to contain spaces.
Also, it might be good to use the long options rather than short for clarity.

tools/airgap_linter.py
porridge: ?

akirillov: @porridge which upstream is meant specifically? if it is about `dcos-commons` then there's a PR https://github.com/mesosphere/dcos-commons/pull/2737

history/universe/marathon.json.mustache
akirillov: done

PR #443, opened 2018-10-27
[DCOS-44204] Fix spark-terasort soak job

samvantran:
## What changes are proposed in this PR? 

Resolves [DCOS-44204](https://jira.mesosphere.com/browse/DCOS-44204)

This PR: 
- adds `user:nobody` to the metronome config for strict clusters. They were immediately failing when set to `root`
- adds the missing cli command `dcos spark`
- always kills the kafka producer (there were times where consumer would crash/test would fail leaving an infinitely streaming producer)

**Note**: this test is run in conjunction with the changes required in the metronome config tracked [here](https://github.com/mesosphere/soak-cluster-configurations/pull/377)

## How was this patch tested?
The metronome job `spark-terasort` on all soak clusters now successfully complete


soumasish:
One general observation though, I saw urllib3 as a dependency in the requirements.txt. If it's not a transitive dependency then we should refactor the code to use requests instead of urllib.
https://www.reddit.com/r/Python/comments/4wl029/requests_vs_urllib_what_problem_does_it_solve/

samvantran:
@soumasish, I would say refactoring is a concern for a separate PR as it seems a bit involved. Perhaps something you'd be interested in undertaking?

mpereira:
I've uploaded the logs for the failed soak111s job to the JIRA ticket. It also looks like the Spark test tooling for creating the diagnostics bundles on test failures didn't get an [upstream fix](https://github.com/mesosphere/dcos-commons/pull/2604). I'll open a PR for that.

mpereira:
https://github.com/mesosphere/spark-build/pull/445

samvantran:
Thanks for that @mpereira. I took a look - the issue looks like it was due to quota. With 3 spark jobs triggering at the same time every hour, it looks like all jobs were contending for resources and were capped by quota (was currently set at 6 cpus).

I looked at the dispatcher UI and it showed some 30+ queued jobs. This was because the kafka soak jobs failed to kill the producers. Specifically, the consumer timed out resulting in the producer staying alive (not getting killed). This happened twice (2 producers constantly running) and this resulted in every subsequent job failing due to not launching.

I removed quota and re-ran the sole spark-terasort job and it passed again. I'll create a separate ticket to fix the spark-kerberos-kafka job so that it always uninstalls the producer.

samvantran:
@elezar I've deleted the requirements file and fixed up the metronome config in the other PR. Please take a 2nd look.

tests/test_soak.py
```diff
@@ -101,6 +103,7 @@
 
 
 def setup_module(module):
+    sdk_cmd.run_raw_cli("package install {} --yes --cli".format(utils.SPARK_PACKAGE_NAME))
```
elezar: Does the `.require` call below not also install the CLI?

tests/test_soak.py
```diff
@@ -101,6 +103,7 @@
...
     if not shakedown.package_installed('spark', SOAK_SPARK_SERVICE_NAME):
         additional_options = {
             "hdfs": {
```
elezar: The hdfs config URI is incorrect for the soak cluster where we install HDFS at `/data-services/hdfs`.

elezar: The `config-url` is incorrect for the soak clusters where HDFS is unstalled under `/data-services/hdfs`.

tests/test_soak.py
```diff
@@ -38,7 +38,9 @@
                  "--keytab-secret-path", "/{}".format(HDFS_KEYTAB_SECRET),
                  "--conf", "spark.mesos.driverEnv.SPARK_USER=root"] # run as root on soak (centos)
```
soumasish: Just a better practise in python land, using an integer explicitly to unpack the format tuple.
```"--keytab-secret-path", "/{0}".format(HDFS_KEYTAB_SECRET),```

Comment (soumasish) [reply tests/test_soak.py]: Otherwise LGTM

Comment (soumasish) [reply tests/test_soak.py]: One general observation though, the sdk dependencies we are pulling in, shouldn't they be refactored as libraries and stay in JFrog and then pulled in via pip?

Comment (soumasish) [reply tests/test_soak.py]: Also one last observation, since the integration tests are all written in python it might be worthwhile to have a python module that wraps the dcos cli commands and can be called as functions. This can also be a pipable dependency.

File: tests/test_soak.py [hunk 4]
```diff
@@ -179,6 +182,10 @@ def _delete_hdfs_terasort_files():
         job_dict = TERASORT_DELETE_JOB_KERBEROS
     else:
         job_dict = TERASORT_DELETE_JOB
+
+    if os.getenv("SECURITY") == "strict":
+        job_dict["run"]["user"] = "nobody"
```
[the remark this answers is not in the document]
elezar: I would rather change the job definition to use `nobody` (as I would assume that this works on permissive as well).

tests/test_soak.py
elezar: > One general observation though, the sdk dependencies we are pulling in, shouldn't they be refactored as libraries and stay in JFrog and then pulled in via pip?

This is what we were discussing in our team meetings last week -- especially the ones related to splitting the repo and consuming the SDK resources.

> Also one last observation, since the integration tests are all written in python it might be worthwhile to have a python module that wraps the dcos cli commands and can be called as functions. This can also be a pipable dependency.

There was a dcos cli python library, but it was deprecated and is unsupported. We have found it to be somewhat more stable to invoke the cli directly.

PR #446, opened 2018-10-31
Added an env and added it to gitignore.

soumasish:
Added an env to build locally in PyCharm. Can be ignored by most thus added to gitignore.


soumasish:
Added an env to build locally in PyCharm. Can be ignored by most thus added to gitignore.

.gitignore
```diff
@@ -27,3 +27,5 @@ spark-universe-url*
 config.yaml
 cluster_info.json
 dcos-launch-*
+env/
```
samvantran: pycharm creates a venv at the root / do you need it at root? We currently have ignores for the same under `/tests`: https://github.com/mesosphere/spark-build/blob/237c6435036aeba5e3bf07969920872775781aeb/.gitignore#L8

.gitignore
```diff
@@ -27,3 +27,5 @@ spark-universe-url*
 config.yaml
 cluster_info.json
 dcos-launch-*
+env/*
+venv/*
```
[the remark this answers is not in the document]
soumasish: @akirillov done.

.gitignore
soumasish: @samvantran This is more of a pycharm requirement, it looks for a env or venv at the root of the project. Also there are .py files in other modules at the same level as tests, so it might be better to have the env at the root and not inside a module.

PR #447, opened 2018-11-06
Integration test for checking to fix to Spark REST submit api.

soumasish:
WIP

### DIFF
--- a/tests/test_spark.py
+++ b/tests/test_spark.py
@@ -503,10 +503,5 @@ def test_handling_wrong_request_to_spark_dispatcher():
             }
     response = requests.post('http://{0}:{1}/submissions/create'.format(dispatcher_ip, port), headers=headers,
                              data=data)
-
     assert (response.status_code < 200 or response.status_code >= 300)
 
-    #check the submission id of the first job to ensure that it's complete to ensure park dispatcher is still running.
-    sdk_cmd.run_raw_cli("dcos spark status {0}".format(submission_id))
-
-    utils.teardown_spark(service_name=service_name)
--- a/tests/test_spark.py
+++ b/tests/test_spark.py
@@ -452,11 +452,11 @@ def test_handling_wrong_request_to_spark_dispatcher():
     headers = {
         'Content-Type': 'application/json;charset=UTF-8',
     }
-    data = {"appResource": "https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar",
+    data = {"action": "CreateSubmissionRequest",
+            "clientSparkVersion": "2.3.2",
+            "appResource": "https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar",
             "sparkProperties": {
-                "spark.master": "spark://{0}".format(host),
-                "spark.driver.cores": "30",
-                "spark.app.name": "SparkPi",
+                "spark.app.name": "TestDispatcher",
 
             },
             "mainClass": "org.apache.spark.examples.SparkPi",
@@ -468,6 +468,5 @@ def test_handling_wrong_request_to_spark_dispatcher():
 
     #check the submission id of the first job to ensure that it's complete to ensure park dispatcher is still running.
     sdk_cmd.run_raw_cli("dcos spark status {0}".format(submission_id))
-
 
     utils.teardown_spark(service_name=service_name)
--- a/tests/test_spark.py
+++ b/tests/test_spark.py
@@ -24,6 +24,7 @@ import sdk_utils
 import requests
 import spark_s3 as s3
 import spark_utils as utils
+import subprocess
 
 LOGGER = logging.getLogger(__name__)
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
@@ -437,14 +438,21 @@ def test_task_stdout():
 @pytest.mark.sanity
 def test_handling_wrong_request_to_spark_dispatcher():
     #submit a correct job
-    #get it's submission id
-    #submit a job through the REST interface with incorrect appArgs
-    #get the submission id, check that it's failed
-    #check the submission id of the first job to ensure that it's complete to ensure park dispatcher is still running.
     service_name = utils.FOLDERED_SPARK_SERVICE_NAME
     utils.require_spark(service_name=service_name)
-    sdk_cmd.run_cli("node ssh --master-proxy --leader")
-    host = "$(hostname)"
+    results  = sdk_cmd.run_raw_cli("dcos spark run --submit-args=\"--class org.apache.spark.examples.SparkPi https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar 30\"")
+
+    #get it's submission id
+    submission_id = results.stdout.split(" ")[-1]
+
+    #submit a job through the REST interface with incorrect appArgs
+
+    #get the submission id, check that it's failed
+    #check the submission id of the first job to ensure that it's complete to ensure park dispatcher is still running.
+
+    sdk_cmd.run_raw_cli("node ssh --master-proxy --leader")
+    results = subprocess.run("hostname", shell=True, universal_newlines=True, check=True)
+    host = results.stdout
     headers = {
         'Content-Type': 'application/json;charset=UTF-8',
     }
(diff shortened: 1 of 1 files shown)


tests/test_spark.py
```diff
@@ -430,3 +433,41 @@ def test_task_stdout():
...
+    service_name = utils.FOLDERED_SPARK_SERVICE_NAME
+    utils.require_spark(service_name=service_name)
+    results  = sdk_cmd.run_raw_cli("dcos spark run --submit-args=\"--class org.apache.spark.examples.SparkPi https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar 30\"")
+
+    #get it's submission id
+    submission_id = results.stdout.split(" ")[-1]
+
+    #submit a job through the REST interface with incorrect appArgs
+    sdk_cmd.run_raw_cli("node ssh --master-proxy --leader")
+    results = subprocess.run("hostname", shell=True, universal_newlines=True, check=True)
```
akirillov: not sure if this command will be actually executed on the `leader` host. In any case, we need to send REST request to Spark Dispatcher, not Mesos Master host.

tests/test_spark.py
```diff
@@ -430,3 +433,41 @@ def test_task_stdout():
...
+    host = results.stdout
+    headers = {
+        'Content-Type': 'application/json;charset=UTF-8',
+    }
+    data = {"appResource": "https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar",
```
akirillov: we don't need to run actual Spark Job, what we need to check is that Dispatcher accepts valid JSON request and _attempts to start_ a job, we are testing Dispatcher liveness in case of JSON request with missing fields.

Valid request:
```
{
    "action": "CreateSubmissionRequest",
    "clientSparkVersion": "1.2.3",
    "appResource": "noop-test.jar",
    "mainClass": "noop.Test",
    "appArgs": ["arg1", "arg2"],
    "environmentVariables": {
        "PATH": "/dev/null"
    },
    "sparkProperties": {
        "spark.app.name": "TestDispatcher"
    }
}
```

Invalid request (no *appArgs* field):
```
{
    "action": "CreateSubmissionRequest",
    "clientSparkVersion": "1.2.3",
    "appResource": "noop-test.jar",
    "mainClass": "noop.Test",
    "environmentVariables": {
        "PATH": "/dev/null"
    },
    "sparkProperties": {
        "spark.app.name": "TestDispatcher"
    }
}
```

Test ahould assert that after submitting an invalid request, the Dispatcher under test is able to accept valid request and _attempts to start a job_

tests/test_spark.py
```diff
@@ -430,3 +433,41 @@ def test_task_stdout():
...
+            "sparkProperties": {
+                "spark.master": "spark://{0}".format(host),
+                "spark.driver.cores": "30",
+                "spark.app.name": "SparkPi",
+
+            },
+            "mainClass": "org.apache.spark.examples.SparkPi",
+            }
+    response = requests.post('http://{}/submissions/create'.format(host), headers=headers, data=data)
```
akirillov: Using `dcos task --json` identify the host and port of the dispatcher. Fields too look at:
- `statuses[0].container_status.network_infos[0].ip_addresses[0].ip_address` - private ip address of the Dispatcher to send REST request to
- `discovery.ports[0].number` - port number to send REST request to

tests/test_spark.py
akirillov: REST request should be sent from a node which has access to private IP address of the Dispatcher, e.g. from a master node using `dcos node ssh --master-proxy --leader "<curl command goes here>"` or by using any of suitable SDK tools

soumasish: Changed testing strategy to pass a valid request, then an invalid request and finally a valid request and check for the status code to be between 200 to 300.

tests/test_spark.py
```diff
@@ -430,3 +433,41 @@ def test_task_stdout():
...
     finally:
         sdk_install.uninstall(utils.SPARK_PACKAGE_NAME, service_name)
+
+
+@pytest.mark.sanity
+def test_handling_wrong_request_to_spark_dispatcher():
+    #submit a correct job
+    service_name = utils.FOLDERED_SPARK_SERVICE_NAME
+    utils.require_spark(service_name=service_name)
+    results  = sdk_cmd.run_raw_cli("dcos spark run --submit-args=\"--class org.apache.spark.examples.SparkPi https://downloads.mesosphere.com/spark/assets/spark-examples_2.11-2.0.1.jar 30\"")
```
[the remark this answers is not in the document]
soumasish: Removed running sparkPi.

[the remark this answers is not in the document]
soumasish: Changed.

tests/test_spark.py
soumasish: Passed the two requests as Python requests object.

tests/test_spark.py
soumasish: Modified approach.

tests/test_spark.py
```diff
@@ -430,3 +433,75 @@ def test_task_stdout():
...
     finally:
         sdk_install.uninstall(utils.SPARK_PACKAGE_NAME, service_name)
+
+
+@pytest.mark.sanity
+def test_handling_wrong_request_to_spark_dispatcher():
+    service_name = utils.FOLDERED_SPARK_SERVICE_NAME
+    utils.require_spark(service_name=service_name)
+
+    json_data = sdk_cmd.run_cli("dcos task --json")
```
akirillov: when a test suite is running, there's going to be multiple tasks. `dcos task --json` will return an array of tasks in JSON format and a spark task can be found in it by name field. After that, the parsing below should work.

tests/test_spark.py
```diff
@@ -430,3 +433,75 @@ def test_task_stdout():
...
+                "PATH": "/dev/null"
+            },
+
+            "sparkProperties": {
+                "spark.app.name": "TestDispatcher",
+
+            },
+
+            }
+    response = requests.post('http://{0}:{1}/submissions/create'.format(dispatcher_ip, port), headers=headers, data=data)
```
akirillov: `requests.post` will execute a HTTP request from the machine where the test is running. Given that it is not guaranteed that Spark Dispatcher will be running on public IP, this request will fail for private IP address. In order to overcome this, the request should be executed _from within_ DCOS cluster, e.g. by using `node ssh "curl..."` or similar approach.

tests/test_spark.py
```diff
@@ -430,3 +433,75 @@ def test_task_stdout():
...
+
+    headers = {
```
akirillov: duplicate definition, variable with the same name and value is already defined in scope

tests/test_spark.py
```diff
@@ -430,3 +433,75 @@ def test_task_stdout():
...
+            },
+
+            }
+    response = requests.post('http://{0}:{1}/submissions/create'.format(dispatcher_ip, port), headers=headers,
+                             data=data)
+
+    headers = {
+        'Content-Type': 'application/json;charset=UTF-8',
+    }
+    data = {"action": "CreateSubmissionRequest",
```
akirillov: it makes sense to define `validRequest` and `invalidRequest` or similar variables to avoid code duplication

tests/test_spark.py
```diff
@@ -430,3 +433,75 @@ def test_task_stdout():
...
+
+            "sparkProperties": {
+                "spark.app.name": "TestDispatcher",
+
+            },
+
+            }
+    response = requests.post('http://{0}:{1}/submissions/create'.format(dispatcher_ip, port), headers=headers,
+                             data=data)
+    assert (response.status_code < 200 or response.status_code >= 300)
```
akirillov: a stronger assertion would be that there are two tasks named `TestDispatcher` present in `dcos task --completed`

tests/test_spark.py
samvantran: We have handy ways to run cmds on masters/agents, see: https://github.com/mesosphere/spark-build/blob/master/testing/sdk_cmd.py#L268

tests/test_spark.py
samvantran: +1, a `validRequest` looks like it's just an additional key-value add to the `invalidRequest`

akirillov:
closing as the author will not be able to finish it

PR #451, opened 2018-11-13
[DCOS-44859] Add parameter files for MWT #10.

mpereira:
The full test parameters are the same from MWT #9.


samvantran:
Until we move all appropriate files to the MWT repo, perhaps we should add all these MWT-related commits to a branch (eg `mixed-workload`) instead of master?

scale-tests/configs/2018-11-15-mwst10.env
```diff
@@ -0,0 +1,145 @@
...
+#   - 1000000 mem
+#   - 0 gpus
+# - executor:
+#   - 0 cpus
+#   - 0 mem
+#   - 0 gpus
+
+# Common streaming jobs configuration ##########################################
+
+TEST_ASSEMBLY_JAR_URL='http://infinity-artifacts.s3.amazonaws.com/scale-tests/dcos-spark-scala-tests-assembly-20180612-5fa9420.jar'
```
samvantran: I don't think anything's changed in the source code since June but if yes then this will change depending on how far we get w/ updating all the latest SHAs, yeah?

mpereira:
> @samvantran: Until we move all appropriate files to the MWT repo, perhaps we should add all these MWT-related commits to a branch (eg mixed-workload) instead of master?

This would add a bit of work to be done wrt updating runbooks and tooling (and possibly some confusion for operators during the next test). Since we should be doing the move soon anyway, I guess we can keep it on master even though it unfortunately kinda pollutes the repo.

PR #456, opened 2018-11-15
SHA-based base image, 'nobody' user, and private keys cleanup for Docker image

akirillov:

Resolves [DCOS-45167](https://jira.mesosphere.com/browse/DCOS-45167)

- switch to SHA instead of a tag for the base image
- user `nobody` is added to Spark Docker image
- `/usr/share/doc` is removed fully because it contains private keys which fail twistlock checks
- multiple similar `RUN` steps are combined to minimize the amount of layers in the image

Integration tests from [mesosphere/spark-build](https://github.com/mesosphere/spark-build)

## Release Notes
- switch to SHA instead of a tag for the base Ubuntu 18.04 image
- user `nobody` is added to Spark Docker image
- unused/example private keys causing Twistlock checks failures removed
- Docker image is optimized to contain fewer layers


elezar:
I see a lot of messages such as:
```
+ + mkdir -p /mnt/mesos/sandbox/nginx
mkdir: cannot create directory '/mnt/mesos/sandbox': Permission denied
+ exec
mkdir -p /mnt/mesos/sandbox/spark
mkdir: cannot create directory '/mnt/mesos/sandbox': Permission denied
```
in the Spark logs. Could it be that we haven't set up the permissions of the sandbox correctly? Is the sandbox now owned by the task user?

universe/config.json
```diff
@@ -40,7 +40,7 @@
                 "user": {
                     "description": "Executors will run as this user.",
                     "type": "string",
-                    "default": "root"
+                    "default": "nobody"
```
samvantran: might want to change it here too: https://github.com/mesosphere/spark-build/blob/master/history/universe/config.json#L33

akirillov:
@elezar the reason is in `nobody`'s UID mismatch between a container and the host OS, working on the fix

universe/marathon.json.mustache
```diff
@@ -106,11 +106,21 @@
...
-        "value": "{{service.user}}"
-        }
+            {
+            "key": "user",
+            "value": "{{service.user}}"
+            }
+        ],
+    {{/service.user}}
+    {{^service.user}}
+        {{#service.OS_flavor}}
```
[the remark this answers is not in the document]
vishnu2kmohan: >  don't quite see the point of using IDs here. Would specifying the nobody user not automatically select 99 on RHEL-based systems? (and 65534 on non-RHEL-based systems)

Yeah the UCR has no problems parsing/translating `nobody` into the appropriate UIDs that match the Host OS but when we run things under the Docker Engine we must explicitly pass the UID (`99` on RHEL/CentOS and `65534` on all the other sane distros) as the `user` param's `value` - It's a known _feature_ :smile:

akirillov:
@elezar thanks for your comments. This PR is under active development right now and more changes coming. `minDcosReleaseVersion` will be `1.10` as you advised. Flag name is updated, thanks for pointing out.

universe/marathon.json.mustache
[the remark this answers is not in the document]
vishnu2kmohan: > This field name looks incorrect

Yeah, it _should_ be: [`service.RHEL_family`](https://github.com/mesosphere/spark-build/blob/baafb30364e0c46c70abb593636e7c44425e9d05/universe/config.json#L69) instead of `service.OS_flavor`

testing/sdk_auth.py
```diff
@@ -30,7 +30,7 @@
 
 KERBEROS_APP_ID = "kdc"
 DCOS_BASE64_PREFIX = "__dcos_base64__"
-LINUX_USER = "core"
+LINUX_USER = os.getenv("DCOS_SSH_USERNAME", "core")
```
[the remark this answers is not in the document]
elezar: I don't see this used anywhere in this file. Is it required here? Looking at the code, the only place where this is used is in `test_hdfs.py` which should be able to user the module variable in `sdk_cmd.py` which reads the same envvar.

testing/sdk_cmd.py
```diff
@@ -22,7 +22,7 @@
 log = logging.getLogger(__name__)
 
 DEFAULT_TIMEOUT_SECONDS = 30 * 60
-
+LINUX_USER = os.getenv("DCOS_SSH_USERNAME", "core")
```
elezar: Same comment as for `sdk_auth` -- specifically about **only** specifying it here.

testing/sdk_install.py
```diff
@@ -147,7 +147,7 @@ def run_janitor(service_name, role, service_account, znode):
 
     auth_token = sdk_cmd.run_cli('config show core.dcos_acs_token', print_output=False).strip()
 
-    cmd_list = ["docker", "run", "mesosphere/janitor", "/janitor.py",
+    cmd_list = ["sudo", "docker", "run", "mesosphere/janitor", "/janitor.py",
```
elezar: Similar note about making sure that this is ported to `dcos-commons`.

spark-testing/spark_utils.py
```diff
@@ -27,7 +27,11 @@
 SPARK_SERVICE_ACCOUNT_SECRET = os.getenv("SPARK_SERVICE_ACCOUNT_SECRET", "spark-service-acct-secret")
 SPARK_SERVICE_NAME = os.getenv("SPARK_SERVICE_NAME", "spark")
 FOLDERED_SPARK_SERVICE_NAME = "/path/to/" + SPARK_SERVICE_NAME
+
 SPARK_USER = os.getenv("SPARK_USER", "nobody")
+SPARK_HISTORY_USER = os.getenv("SPARK_USER", SPARK_USER)
```
elezar: There is never a way that these could differ (unless they are overwritten after executing these lines):
```suggestion
SPARK_HISTORY_USER = SPARK_USER
```

tests/test_kafka.py
```diff
@@ -103,6 +103,7 @@ def setup_spark(kerberized_kafka, configure_security_spark, configure_universe):
     try:
         utils.upload_dcos_test_jar()
         utils.require_spark()
+        sdk_cmd.run_cli('package install --cli dcos-enterprise-cli --yes')
```
elezar: It is very strange to me that the installation of `kerberized_kafka` does not install this CLI as it is **required** there to create the secrets storing the Kerberos keytabs.
```suggestion
```

tests/test_spark.py
```diff
@@ -321,7 +322,6 @@ def test_s3_env():
 def test_foldered_spark():
     service_name = utils.FOLDERED_SPARK_SERVICE_NAME
     zk = 'spark_mesos_dispatcher__path_to_spark'
-    utils.teardown_spark(service_name=service_name, zk=zk)
```
elezar: This is an optimisation to not have spark uninstalled and installed all the time, correct?

universe/marathon.json.mustache
```diff
@@ -104,20 +104,40 @@
...
+                "key": "user",
+                "value": "{{service.docker_user}}"
+            }
+        ],
+    {{/service.docker_user}}
+    {{^service.docker_user}}
+        "parameters": [
+            {
+                "key": "user",
+                "value": "{{service.user}}"
```
elezar: Question: Is `service.user` required? If not, will these parameters work as expected if it is not supplied? I'm not saying we should make the logic here more complex, I'm just asking.

universe/config.json
```diff
@@ -71,6 +71,11 @@
                     "description": "Launch the Dispatcher using the Universal Container Runtime (UCR)",
                     "default": false
                 },
+                "docker_user": {
```
elezar: Should we add a note that this is *only* required if it **differs** from `service.user`?

tests/test_hdfs.py
```diff
@@ -269,13 +268,13 @@ def test_supervise():
         stop_max_delay=600*1000,
         retry_on_result=lambda res: not res)
     def wait_job_present(present):
-        svc = shakedown.get_service(JOB_SERVICE_NAME)
+        svc = shakedown.get_service(job_service_name)
```
elezar: Nit: I would prefer to have `job_service_name` be a parameter for this inner function, or move the declaration of the `job_service_name` to before this function to make the intent clearer.

tests/test_spark.py
akirillov: not exactly, the next line after this one calls method `utils.require_spark` which includes teardown as a part of it.

universe/marathon.json.mustache
akirillov: service user comes here from `config.json` and if not specified is set to default value of `nobody`.  one of the interesting findings was that if docker user parameter not supplied at all - then container is run as root even if Marathon service user is different.

Here is why `service.user` is alway present at the moment of template evaluation:
```
Variables in the mustache template will be evaluated from a union object created by merging three objects in the following order:
- Defaults specified in config.json
- User supplied options from either the DC/OS CLI or the DC/OS UI
- The contents of resource.json
```

universe/config.json
akirillov: I updated description, please check if it reflects the purpose of this property clearly.

tests/test_hdfs.py
akirillov: fixed

tests/test_kafka.py
akirillov: removed

docker/Dockerfile
```diff
@@ -17,40 +17,43 @@
...
-    cd /opt/mesosphere && \
-    curl -L -O https://downloads.mesosphere.io/libmesos-bundle/libmesos-bundle-1.12.0.tar.gz && \
-    tar zxf libmesos-bundle-1.12.0.tar.gz && \
-    rm libmesos-bundle-1.12.0.tar.gz && \
-    curl -L -O https://downloads.mesosphere.com/dcos-commons/artifacts/0.50.0/bootstrap.zip && \
-    unzip bootstrap.zip && \
-    rm bootstrap.zip
+RUN mkdir -p /opt/mesosphere/ \
+    && cd /opt/mesosphere \
+    && curl -L -O https://downloads.mesosphere.io/libmesos-bundle/libmesos-bundle-1.12.0.tar.gz \
```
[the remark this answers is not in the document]
akirillov: these parts are refactored now to be more DRY

testing/sdk_auth.py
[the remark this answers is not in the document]
akirillov: removed from here to `sdk_cmd.py`

spark-testing/spark_utils.py
akirillov: good catch, fixed

universe/marathon.json.mustache
```diff
@@ -104,20 +104,40 @@
...
         ],
-    {{/service.user}}
+    {{/service.docker_user}}
     "forcePullImage": true
     }
     },
 {{/service.UCR_containerizer}}
 
     "healthChecks": [
+    {{#service.UCR_containerizer}}
```
samvantran: 👍

testing/sdk_install.py
akirillov: done: https://github.com/mesosphere/dcos-commons/pull/2793

spark-testing/spark_utils.py
```diff
@@ -129,6 +137,9 @@ def submit_job(
 
     conf_args += ['--conf', 'spark.mesos.role={}'.format(driver_role)]
 
+    if SPARK_DOCKER_USER is not None:
+        conf_args += ['--conf', 'spark.mesos.executor.docker.parameters=user={}'.format(SPARK_DOCKER_USER)]
```
mpereira: I noticed this line isn't being exercised by the integration tests. I'm just curious, do you know if some sort of quoting is required here due to the necessity of `=` characters in the value for `spark.mesos.executor.docker.parameters`? e.g. `spark.mesos.executor.docker.parameters="user=root"`

akirillov: this is not required although it's not quite reflected in the [docs](https://spark.apache.org/docs/2.2.0/running-on-mesos.html) but the source code here: https://github.com/mesosphere/spark/blob/master/resource-managers/mesos/src/main/scala/org/apache/spark/scheduler/cluster/mesos/MesosSchedulerBackendUtil.scala#L150-L153 and here: https://github.com/mesosphere/spark/blob/master/resource-managers/mesos/src/main/scala/org/apache/spark/scheduler/cluster/mesos/MesosSchedulerBackendUtil.scala#L111-L127 proves that quotes are not needed.

mpereira: Sweet!

PR #458, opened 2018-11-28
[DCOS-45356] Add retry to curl calls in test_unique_vips

samvantran:

Resolves [DCOS-45356](https://jira.mesosphere.com/browse/DCOS-45356)

Test appears to be flaking in nightlies and occasional PRs. From logs, both dispatchers are deployed healthily but perhaps it takes a few seconds for the DNS entries to prop. This PR adds retries for a little bit of resilience.

Integration test passes


tests/test_spark.py
```diff
@@ -387,21 +387,23 @@ def test_driver_executor_tls():
 
 @pytest.mark.sanity
 def test_unique_vips():
+
+    @retrying.retry(stop_max_attempt_number=3, wait_fixed=2000)
+    def verify_ip_is_reachable(ip):
+        ok, _ = sdk_cmd.master_ssh("curl {}".format(ip))
+        assert ok
```
akirillov: what will happen if the assertion fails? is this type of error also handled by `retry`?

samvantran: Yup `retry` catches any exception that gets thrown and will retry based on params

PR #459, opened 2018-11-28
Spark Structured Streaming checkpointing integration test

akirillov:

Resolves [DCOS-41580](https://jira.mesosphere.com/browse/DCOS-41580)

The new integration test launches Spark Structured Streaming job with checkpointing backed by HDFS, feeds some test data to Kafka and verifies that the aggregated data is not reprocessed but recovered from a checkpoint instead if a failure happens. All components are Kerberized.

Apart from the test this PR introduces a refactoring for KDC/HDFS/Kafka-dependent tests and allows reuse of these components across the tests without reinstalling them every time. 

Changes:
- Spark structured streaming checkpointing recovery integration test
- session-scoped shared KDC, HDFS, and Kafka services
- separate service accounts and secrets for Kafka and HDFS
- shared KDC, HDFS, and Kafka configuration and tooling was extracted to 'fixtures' modules
- Kerberized Kafka Feeder and Spark Structured Streaming test jobs
- Spark version upgrade to 2.3.2

Integration tests from this repo

## Release Notes

* Upgrade of Spark version to 2.3.2
* Structured Streaming checkpointing and recovery integration test


tests/fixtures/__init__.py
```diff
@@ -0,0 +1,7 @@
+import os.path
+import sys
+
+this_file_dir = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.normpath(os.path.join(this_file_dir, '..', '..', 'tests')))
+sys.path.append(os.path.normpath(os.path.join(this_file_dir, '..', '..', 'testing')))
+sys.path.append(os.path.normpath(os.path.join(this_file_dir, '..', '..', 'spark-testing')))
```
elezar: Nit: Newline:
```suggestion
sys.path.append(os.path.normpath(os.path.join(this_file_dir, '..', '..', 'spark-testing')))

```

tests/fixtures/hdfs.py
```diff
@@ -0,0 +1,195 @@
...
+
+        sdk_install.uninstall(HDFS_PACKAGE_NAME, HDFS_SERVICE_NAME)
+        sdk_install.install(
+            HDFS_PACKAGE_NAME,
+            HDFS_SERVICE_NAME,
+            DEFAULT_HDFS_TASK_COUNT,
+            additional_options=service_kerberos_options,
+            timeout_seconds=30 * 60)
+
+        yield kerberos_env
```
elezar: Why does this yield `kerberos_env`? This fixture doesn't modify the env at all (that I can tell).

tests/fixtures/hdfs.py
```diff
@@ -0,0 +1,195 @@
...
+
+    finally:
+        sdk_install.uninstall(HDFS_PACKAGE_NAME, HDFS_SERVICE_NAME)
+        kerberos_env.cleanup()
```
elezar: In my opinion, this should be handled in the fixture providing `kerberos_env`.

tests/fixtures/hdfs.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@pytest.fixture(scope='session')
+def setup_hdfs_client(hdfs_with_kerberos):
```
elezar: It is strange that this depends on `hdfs_with_kerberos` without accessing it. Should the client depend on the service in some way?

tests/fixtures/hdfs.py
```diff
@@ -0,0 +1,195 @@
...
+        additional_options = {
+            "hdfs": {
+                "config-url": "http://api.{}.marathon.l4lb.thisdcos.directory/v1/endpoints".format(HDFS_SERVICE_NAME)
+            },
+            "security": {
+                "kerberos": {
+                    "enabled": True,
+                    "realm": sdk_auth.REALM,
+                    "kdc": {
+                        "hostname": hdfs_with_kerberos.get_host(),
```
elezar: I think that we should explicitly use `kerberos_env` here.

tests/fixtures/hdfs.py
```diff
@@ -0,0 +1,195 @@
...
+    finally:
+        sdk_install.uninstall(HISTORY_PACKAGE_NAME, HISTORY_SERVICE_NAME)
+
+
+@pytest.fixture(scope='module')
+def kerberized_spark(setup_history_server, hdfs_with_kerberos, configure_security_spark, configure_universe):
+    try:
+        additional_options = {
+            "hdfs": {
+                "config-url": "http://api.{}.marathon.l4lb.thisdcos.directory/v1/endpoints".format(HDFS_SERVICE_NAME)
```
elezar: Note: I'm not a big fan of these global variables, and would rather access the `hdfs_with_kerberos` to obtain these values. Not going to block on this though because I understand that that would be a larger refactoring.

tests/fixtures/kafka.py
```diff
@@ -0,0 +1,97 @@
...
+
+        sdk_install.uninstall(KAFKA_PACKAGE_NAME, KAFKA_SERVICE_NAME)
+        sdk_install.install(
+            KAFKA_PACKAGE_NAME,
+            KAFKA_SERVICE_NAME,
+            DEFAULT_KAFKA_TASK_COUNT,
+            additional_options=service_options,
+            timeout_seconds=30 * 60)
+
+        yield kerberos_env
```
elezar: Same comment as with HDFS. This should not be `kerberos_env`.

tests/fixtures/kafka.py
```diff
@@ -0,0 +1,97 @@
...
+
+    finally:
+        sdk_install.uninstall(KAFKA_PACKAGE_NAME, KAFKA_SERVICE_NAME)
+        kerberos_env.cleanup()
+        print('noop')
```
elezar:
```suggestion


```

tests/fixtures/kafka.py
```diff
@@ -0,0 +1,97 @@
...
+            KAFKA_SERVICE_NAME,
+            DEFAULT_KAFKA_TASK_COUNT,
+            additional_options=service_options,
+            timeout_seconds=30 * 60)
+
+        yield kerberos_env
+
+    finally:
+        sdk_install.uninstall(KAFKA_PACKAGE_NAME, KAFKA_SERVICE_NAME)
+        kerberos_env.cleanup()
```
elezar: As is the case with HDFS, we should not cleanup the Kerberos environment here.

tests/fixtures/kdc.py
```diff
@@ -0,0 +1,71 @@
...
+        kafka_principals = build_kafka_principals()
+        hdfs_principals = build_hdfs_principals()
+
+        kerberos_env = sdk_auth.KerberosEnvironment(persist=True)
+        kerberos_env.add_principals(kafka_principals + hdfs_principals)
+        kerberos_env.finalize()
+        yield kerberos_env
+
+    finally:
+        print('noop')
```
elezar:
```suggestion
```

tests/test_checkpointing.py
```diff
@@ -0,0 +1,122 @@
...
+        "--conf", "spark.scheduler.maxRegisteredResourcesWaitingTime=2400s",
+        "--conf", "spark.scheduler.minRegisteredResourcesRatio=1.0",
+        "--conf", uris
+    ] + kafka_kerberos_args
+
+    # configuring streaming job and HDFS folders
+    setup_hdfs_paths()
+
+    # running kafka producer
+    messages = ["test"] * 100
```
elezar: Would it make sense to make these messages unique for a specific run?:
`["test0", "test1", ... ]`?

 ```suggestion
     messages = ["test%d" % d for d in range(100)]
```

File: tests/test_soak.py [hunk 12]
```diff
@@ -9,7 +9,7 @@
 import sdk_utils
 
 import spark_utils as utils
-from tests.test_kafka import KAFKA_PACKAGE_NAME, test_pipeline
+# from tests.test_kafka import KAFKA_PACKAGE_NAME, test_pipeline
```
elezar: Let's rather remove this than comment it out.

tests/fixtures/hdfs.py
akirillov: The main intent of this dependency is to introduce the ordering in which `hdfsclient` never created before hdfs service. What would be a better way of implementing this?

tests/test_soak.py
akirillov: this was commented by mistake, reverted back now.

tests/test_checkpointing.py
akirillov: Good catch! The new implementation uses two different word sets to verify both the count accuracy and aggregated data checkpointing.

tests/fixtures/kdc.py
akirillov: this wasn't intended to be committed, removed

tests/fixtures/kafka.py
akirillov: cleanup moved to KDC fixture

tests/fixtures/kafka.py
akirillov: this wasn't intended to be committed, removed

tests/fixtures/kafka.py
akirillov: fixed

tests/fixtures/hdfs.py
akirillov: Totally agree but I'd prefer to not mix necessary refactorings needed for making the test work with general code refactoring even if it is needed. Integration tests codebase definitely needs a few rounds of refactoring so let's address them in a separate PRs.

tests/fixtures/hdfs.py
akirillov: switched to `kerberos_env`

tests/fixtures/hdfs.py
akirillov: cleanup moved to KDC fixture

tests/fixtures/hdfs.py
akirillov: removed in favour of using `kerberos_env` throughout the dependent tests and fixtures.

tests/fixtures/__init__.py
akirillov: added

akirillov:
thanks for your comments @elezar. I cleaned up the code according to your suggestions and left a comment related to a wider refactoring. The Structured Streaming test itself is modified now to avoid false positives by using different word sets fed into Kafka before and after driver failure.

tests/fixtures/hdfs.py
elezar: I realise that it's about ordering, so that's fine, but the dependence is not really explicit here. It takes knowledge of the client itself to know that it **requires** an HDFS service with a name `hdfs`.

What I have used in the past is something like: https://github.com/mesosphere/dcos-commons/blob/d83139ac0fb9c53719a9e5637a5704ae9dfe23d5/frameworks/hdfs/tests/test_kerberos_auth.py#L101

where the `hdfs_with_kerberos` is a `dict` which contains the required configuration and this si explicitly passed to the client. It's not a requirement for this PR though.

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    val Array(service, brokers, topic) = args.take(3)
+    val messages = args.drop(3)
+    println(s"Got brokers $brokers, producing ${messages.length} to topic $topic")
+    try {
+      withProducer(service, brokers, sendMessages(topic, messages))
+    } catch {
+      case t: Throwable =>
+        t.printStackTrace()
+        Thread.sleep(30*60*1000)
```
samvantran: why sleep for 30 minutes?

tests/test_checkpointing.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @retrying.retry(
+        wait_fixed=5000,
+        stop_max_delay=600 * 1000,
+        retry_on_result=lambda res: not res)
+    def wait_for_running_job_output(task_id, expected_line):
+        stdout = sdk_cmd.run_cli("task log --lines=1000 {}".format(task_id))
+        result = expected_line in stdout
+        LOGGER.info('Checking for {} in STDOUT:\n{}\nResult: {}'.format(expected_line, stdout, result))
+        return result
```
samvantran: This looks mostly like [spark_utils.check_job_output](https://github.com/mesosphere/spark-build/blob/master/spark-testing/spark_utils.py#L164) with some retries. Can you move this method to that class or extend the code?

tests/test_checkpointing.py
```diff
@@ -0,0 +1,126 @@
...
+    application_args = "{} {} {} {}".format(kafka_brokers, KAFKA_TEST_TOPIC, HDFS_CHECKPOINT_DIR, SPARK_SECURITY_PROTOCOL)
+
+    driver_task_id = utils.submit_job(app_url=jar_uri,
+                                      app_args=application_args,
+                                      service_name=utils.SPARK_SERVICE_NAME,
+                                      args=(SPARK_SUBMIT_HDFS_KERBEROS_ARGS + spark_submit_args))
+
+    # Wait until executor is running
+    LOGGER.info("Starting supervised driver {}".format(driver_task_id))
+    sdk_tasks.check_running(SPARK_APPLICATION_NAME, 1, timeout_seconds=600)
```
samvantran: can you add `expected_task_count=1` otherwise the lone `1` is a bit mysterious or use a named_constant because I see something similar on L117.

tests/test_hdfs.py
```diff
@@ -1,238 +1,25 @@
...
-            }
-        }
-        utils.require_spark(additional_options=additional_options)
-        yield
-    finally:
-        utils.teardown_spark()
 
+from tests.fixtures.hdfs import HDFS_SERVICE_NAME, HDFS_DATA_DIR, HDFS_HISTORY_DIR
+from tests.fixtures.hdfs import HISTORY_PACKAGE_NAME
+from tests.fixtures.hdfs import SPARK_SUBMIT_HDFS_KERBEROS_ARGS, KEYTAB_SECRET_PATH, GENERIC_HDFS_USER_PRINCIPAL
```
samvantran: nit: need 2 line breaks b/w imports and code

tests/test_hdfs.py
```diff
@@ -324,20 +111,20 @@ def wait_job_present(present):
 
 @pytest.mark.skipif(not utils.hdfs_enabled(), reason='HDFS_ENABLED is false')
 @pytest.mark.sanity
-def test_history():
+def test_history(kerberized_spark, hdfs_with_kerberos, setup_history_server):
```
samvantran: if you're depending on`kerberized_spark` (which internally depends on `setup_history_sever, hdfs_with_kerberos`), do you need to call those two again here?

tests/test_kafka.py
```diff
@@ -150,21 +67,9 @@ def test_pipeline(kerberos_flag, stop_count, jar_uri, keytab_secret, spark_servi
...
-        "--conf", "spark.mesos.driverEnv.KRB5_CONFIG_BASE64={}".format(KAFKA_KRB5),
-        "--conf", "spark.driver.extraJavaOptions=-Djava.security.auth.login.config="
-                  "/mnt/mesos/sandbox/spark-kafka-client-jaas.conf",
-        "--conf", "spark.executor.extraJavaOptions="
-                  "-Djava.security.auth.login.config=/mnt/mesos/sandbox/spark-kafka-client-jaas.conf",
-    ]
+    kerberos_args = get_kerberized_kafka_spark_conf(spark_service_name)
 
-    producer_config = ["--conf", "spark.cores.max=2", "--conf", "spark.executor.cores=2",
+    producer_config = ["--conf", "spark.cores.max=2", "--conf", "spark.executor.cores=1",
```
samvantran: This will mean that we spawn 2 producers, is this intended?

tests/test_kafka.py
```diff
@@ -177,7 +82,8 @@ def test_pipeline(kerberos_flag, stop_count, jar_uri, keytab_secret, spark_servi
 
     sdk_tasks.check_running(KAFKA_SERVICE_NAME, 1, timeout_seconds=600)
 
-    consumer_config = ["--conf", "spark.cores.max=4", "--class", "KafkaConsumer"] + common_args
+    consumer_config = ["--conf", "spark.cores.max=2", "--conf", "spark.executor.cores=1",
```
samvantran: 2 producers, 2 consumers - okay, looks like it is intended

manifest.json
```diff
@@ -1,17 +1,17 @@
 {
-    "spark_version": "2.2.1",
+    "spark_version": "2.3.2",
     "default_spark_dist": {
         "hadoop_version": "2.7",
-        "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.2.1-4-SNAPSHOT-bin-2.7.tgz"
+        "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.3.2-1-SNAPSHOT-bin-2.7.3.tgz"
```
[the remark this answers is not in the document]
akirillov: Jenkins upload job is failing if an object with the same key exists, this complicated SNAPSHOT replacement. Overall, I think we should still keep some sort of versioning to avoid failures when a snapshot is overridden with some broken/incompatible changes.

manifest.json
```diff
@@ -1,17 +1,17 @@
...
     },
     "spark_dist": [
         {
             "hadoop_version": "2.7",
-            "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.2.1-4-SNAPSHOT-bin-2.7.tgz"
+            "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.3.2-1-SNAPSHOT-bin-2.7.3.tgz"
```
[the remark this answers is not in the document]
akirillov: There's a separate ticket for porting CVEs commit from branch 2.2 to 2.3, in the current branch Hadoop is of version 2.7.3. It makes sense to (at least) update Hadoop dependency first in 2.3 branch and use the updated tar here to avoid having an outdated master in `spark-build`

tests/fixtures/kafka.py
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        service_options = {
+            "service": {
+                "name": KAFKA_SERVICE_NAME,
+                "security": kerberos_options
+            },
+            "kafka": {
+                "default_replication_factor": 3,
+                "num_partitions": 32
```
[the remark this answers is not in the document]
akirillov: Yes, this unstuck Spark Structured Streaming consumer.  `spark_kafka` test has stabilized too but it looks like the stuck consumption issue is still present there (long freeze before starting actual consumption).

tests/jobs/scala/build.sbt
```diff
@@ -1,4 +1,4 @@
-lazy val SparkVersion = sys.env.get("SPARK_VERSION").getOrElse("2.2.0")
+lazy val SparkVersion = sys.env.getOrElse("SPARK_VERSION", "2.3.2")
```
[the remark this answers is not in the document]
akirillov: thanks for this catch! will add to gitignore and cleanup the PR from it

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
akirillov: This was added to ease Kafka producer troubleshooting in case of errors (happens only when an exception is thrown). After verifying functionality it looks like it's not needed anymore.

tests/test_checkpointing.py
akirillov: good catch. fixed that.

tests/test_kafka.py
akirillov: yes, the main goal here is to lower down a single executor resource demands while keeping parallelism at the same level.

tests/test_kafka.py
akirillov: yes, it is the same here - fixed-size executors preserving specified parallelism level.

tests/test_checkpointing.py
akirillov: the implementation is a bit different for checking the output of running vs completed tasks so I moved this method to `spark_utils`

tests/test_hdfs.py
samvantran: okay that's fair

tests/test_checkpointing.py
samvantran: okay

manifest.json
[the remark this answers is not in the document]
samvantran: Okay let's port the CVE updates first and rebase this

manifest.json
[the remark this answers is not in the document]
samvantran: If this is the case then do we need the word `SNAPSHOT` at all? We can go with a simpler versioning system like `2.3.2.1` or `2.3.2-1`, if we make it known the last digit represents a dev build. Or perhaps a proper thing to do is tag the file w/ a date-time

PR #467, opened 2019-01-04
Long-running Spark Application for testing and troubleshooting

akirillov:

A test Spark application for emulating long-running jobs like Spark Streaming for triaging and troubleshooting purposes

Manual execution on a test DCOS cluster

## Release Notes

N/A


samvantran:
I'm not really against adding this app bc its intent is clear but FYI we do have the [monte carlo app](https://github.com/mesosphere/spark-build/blob/master/scale-tests/apps/monte-carlo-portfolio.py) that we use for MWTs that we can make long running by passing bigger inputs for `scale_factor`

akirillov:
@samvantran, yes, we can also use SparkPi with a larger input, or any other one. The main goal of this specific app is that it is deterministic and the user explicitly specifies the duration of each task, not the number of iterations.

tests/jobs/scala/src/main/scala/TaskRunnerApp.scala
```diff
@@ -0,0 +1,33 @@
...
+    val spark = SparkSession
+      .builder
+      .appName("TaskRunnerApp")
+      .getOrCreate()
+
+    val numTasks = args(0).toInt
+    val sleepMillis = args(1).toInt * 1000
+
+    spark.sparkContext.parallelize(0 until numTasks, numTasks).foreachPartition{_ =>
+      println(s"Sleeping for $sleepMillis")
```
samvantran: can you append `s` or `seconds` after `$sleepMillis`?

tests/jobs/scala/src/main/scala/TaskRunnerApp.scala
```diff
@@ -0,0 +1,33 @@
+import org.apache.spark.sql.SparkSession
+
+/**
+  * TaskRunnerApp creates a number of noop Spark tasks specified by 'numTasks' each of which
+  * sleeps for 'taskDurationSec'. The main goal of this class is to mimic long-running Spark
+  * applications (i.e. Spark Streaming) for the testing and troubleshooting purposes.
```
mpereira: LGTM. Only one nit: if the intention for this class is to generate "sleep tasks", should we rename it accordingly to be more precise?

tests/jobs/scala/src/main/scala/TaskRunnerApp.scala
akirillov: added

tests/jobs/scala/src/main/scala/TaskRunnerApp.scala
akirillov: renamed to `MockTaskRunner` in order to avoid implementation exposure so it can be changed to something else instead of `Thread.sleep()`

PR #478, opened 2019-01-22
[DCOS-47567] Fix failing CNI tests for DCOS 1.10 and 1.11

akirillov:

Resolves [DCOS-47567](https://jira.mesosphere.com/browse/DCOS-47567)

With older versions of Mesos `dcos task exec` command produced an additional output which broke IP address parsing in tests. This PR fixes this issue by using only the last line of command's output.

Integration tests from this repo against DCOS 1.10 and 1.11 clusters before and after the fix

## Release Notes

N/A


samvantran:
1.10 and 1.11 strict tests passed so looks like we're all good 👍

PR #483, opened 2019-01-30
[DCOS-47872] Teardown Spark Drivers after tests time out waiting for completion

akirillov:

Resolves [DCOS-47870](https://jira.mesosphere.com/browse/DCOS-47870)

Currently, several Spark soak tests wait for a specific output for a specified amount of time and then terminate with an error if this output is not found leaving Spark Driver running. This leads to quota exhaustion and makes future test jobs fail creating a backlog of timed out tests which never succeed.

This PR adds termination of the timed out drivers in affected tests.

- integration tests from this repo
- soak tests executed from the current PR branch

## Release Notes

N/A


spark-testing/spark_utils.py
```diff
@@ -35,7 +35,9 @@
 SPARK_DOCKER_IMAGE = os.getenv("DOCKER_DIST_IMAGE", "mesosphere/spark-dev")
 
 SPARK_DRIVER_ROLE = os.getenv("SPARK_DRIVER_ROLE", "*")
-JOB_WAIT_TIMEOUT_SEC = 1800
+
```
soumasish: Any particular reason for these numbers?

soumasish:
LGTM

spark-testing/spark_utils.py
```diff
@@ -179,7 +187,7 @@ def submit_job(
 
 def check_job_output(task_id, expected_output):
     LOGGER.info('Waiting for task id={} to complete'.format(task_id))
-    shakedown.wait_for_task_completion(task_id, timeout_sec=JOB_WAIT_TIMEOUT_SEC)
+    shakedown.wait_for_task_completion(task_id, timeout_sec=JOB_WAIT_TIMEOUT_SECONDS)
```
imaxxs: AFAIK, SDK 0.55.x got rid of shakedown library. This may be separate task but we should check if we are using the shakedown in spark tests and create tickets to remove them eventually. You can double check with @takirala  and @kvish about this.

imaxxs:
Other than my comments, LGTM.

spark-testing/spark_utils.py
akirillov: @imaxxs, we should depend on SDK in a different way (not by duplicating the code). Currently, codebases diverged and moving off of shakedown might be painful. Anyway, we have to do it.

PR #487, opened 2019-02-25
[DCOS-48992] Add parameter files for the MWT #13.

mpereira:

Resolves [DCOS-48992](https://jira.mesosphere.com/browse/DCOS-48992)

- Adds parameter files for the MWT #13.

These changes will be tested on soak113s on 2019-02-27.


scale-tests/configs/2019-02-28-mwst13.env
```diff
@@ -0,0 +1,147 @@
...
+BATCH_SPARK_BUILD_BRANCH=master
+
+# Batch GPU jobs configuration #################################################
+
+GPU_DISPATCHERS_JSON_OUTPUT_FILE_URL="https://${TEST_S3_BUCKET}.s3.amazonaws.com/${TEST_S3_FOLDER}/${GPU_DISPATCHERS_JSON_OUTPUT_FILE}"
+
+GPU_APP_ID="/${SERVICE_NAMES_PREFIX}gpu-batch-workload"
+GPU_SCRIPT_CPUS=2
+GPU_SCRIPT_MEM=4096
+GPU_DOCKER_IMAGE='samvantran/spark-dcos-gpu:metrics'
```
samvantran: This image was built on spark 2.2.1 and given that we are now soaking latest spark 2.3.2 and are unable to test running gpu workloads on soak environments, I would say on actual MWT day we should either 1) not run them or 2) deploy gpu dispatchers and run workload after the normal spark workloads are in good shape (eg 250-500 running jobs; no failures). If they run fine, then great! If not, shut em down immediately. WDYT?
cc @vishnu2kmohan

mpereira: Would it work if we used `mesosphere/spark:2.6.0-2.3.2-hadoop-2.7` instead?

By the way, Spark 2.3 is still pending: https://jira.mesosphere.com/browse/DCOS-49000

samvantran: Nah, the spark-dcos-gpu image actually includes installing all the cuda/cudnn libraries + Tensorflow in order to run [this TF app](https://github.com/mesosphere/spark-build/blob/master/scale-tests/batch_test.py#L57)

mpereira: I looked for a bit and couldn't find the Dockerfile for that image. Could you please share it? We should probably get it checked in this repo.

samvantran: ah sorry about that - this is the one: https://github.com/samvantran/spark-dcos-docker/blob/master/Dockerfile-statsD-metrics 
but it also builds from https://github.com/samvantran/spark-dcos-docker/blob/master/Dockerfile-minpatch-nv390-cuda9-cudnn7115-tf15

mpereira: Thanks Sam. I created https://jira.mesosphere.com/browse/DCOS-49094.

mpereira: I had a chat with SRE and they think they'll be able to add a GPU agent to the Soak cluster by tomorrow, so assuming we use the latest Spark version, we'll be able to test whether or not using the "spark-dcos-gpu" (which lower Spark version) works for running jobs against it.

mpereira: Pushed commit taking advantage of the GPU node: https://github.com/mesosphere/spark-build/pull/487/commits/9b45961ebcf3c555334b248c46a2b9a27269b1e7

mpereira: @samvantran on today's test I noticed that GPU jobs were not getting scheduled by Spark: https://soak113s.testing.mesosphe.re/service/data-services-testmwst13/gpu-spark-00/

Maybe it's related to the current Docker image with CUDA libraries not being compatible with the newest Spark version? If you have some time, could you take a look into this @samvantran?

In any case I'll attempt to submit the GPU jobs tomorrow and see what happens. Unless you discover that the Docker image Spark version mismatch is indeed the reason.

samvantran: I don't think it's a matter of the docker image. These jobs just aren't being scheduled. 

My guess is that soak113 is not set up the same as MWST. From `/mesos/flags` I see 
```
"filter_gpu_resources":"true"
fair_sharing_excluded_resource_names":"{ gpus }"
```
which leads me to [MESOS-5377](https://issues.apache.org/jira/browse/MESOS-5377) which seems to affect DRF strategy specifically for what is happening in this cluster, 1 gpu agent vs many non-gpu. 

I would say still worth attempting to run GPU jobs tmrw and if they don't run we can triage why then.

mpereira: Ack. Thanks for taking a look!

PR #492, opened 2019-03-06
[DCOS-49020] User propagation and override support for Spark Dispatcher

akirillov:

Resolves [https://jira.mesosphere.com/browse/DCOS-49020](https://jira.mesosphere.com/browse/DCOS-49020)

* Spark Mesos Dispatcher will launch Driver and Executors using `{{service.user}}` by default
* Tests to verify support for user override via `spark.mesos.executor.docker.parameters=user=<user>` and `spark.mesos.driverEnv.SPARK_USER=<user>`

* newly added integration tests covering all combination of containerizers for Dispatcher and Spark apps
* running `test_spark_users.py` against both CoreOS and CentOS clusters

## Release Notes

* Spark Mesos Dispatcher launches Driver and Executors using `{{service.user}}` by default for both Docker and Mesos containerizers
* Improved support for `docker_user` UID override for CentOS/RHEL (no need to specify UID at the time of job submit)
* Added support for user override for submitted jobs. It possible to launch a job as `nobody` while Dispatcher is running is `root`


docker/runit/service/spark/run
```diff
@@ -76,6 +76,14 @@ function configure_properties() {
         add_if_non_empty spark.mesos.dispatcher.driverDefault.spark.executorEnv.VIRTUAL_NETWORK_ENABLED "${VIRTUAL_NETWORK_ENABLED}"
     fi
 
+    if [ "${SPARK_DOCKER_USER}" != "" ]; then
+        echo "spark.mesos.dispatcher.driverDefault.spark.mesos.executor.docker.parameters=user=${SPARK_DOCKER_USER}" >> ${SPARK_HOME}/conf/mesos-cluster-dispatcher.properties
```
samvantran: wow what a config +1 for taking care of this for the user

tests/test_spark_users.py
```diff
@@ -0,0 +1,137 @@
...
+        yield
+    finally:
+        utils.teardown_spark(service_name=SERVICE_NAME, zk=DISPATCHER_ZK)
+
+
+@pytest.mark.sanity
+@pytest.mark.parametrize('user,use_ucr_containerizer,use_ucr_for_spark_submit', [
+    ("nobody", True, True),
+    ("nobody", True, False),
+    ("nobody", False, True),
```
samvantran: Do we want to test scenarios where users launch dispatchers with docker but submit jobs with UCR? I guess it doesn't hurt but is this realistic?

tests/test_spark_users.py
```diff
@@ -0,0 +1,137 @@
...
+    submit_args = ["--conf spark.cores.max=1",
+                   "--class {}".format(app_name)] + extra_args
+
+    driver_task_id = utils.submit_job(service_name=SERVICE_NAME,
+                                      app_url=utils.dcos_test_jar_url(),
+                                      app_args="1 300",
+                                      args=submit_args)
+    try:
+        sdk_tasks.check_running(app_name, 1, timeout_seconds=300)
+        driver_task = shakedown.get_task(driver_task_id, completed=False)
```
samvantran: we ought to move away from `shakedown` but it doesn't have to be this PR

tests/test_spark_users.py
akirillov: By default, Dispatcher is deployed using Docker containerizer. From here we can think of the next scenarios:
- a user submits a job (`False`, `False`)
- a user discovers UCR option, updates Dispatcher config, but keeps submitting jobs as before (`True`, `False`)
- user discovers spark-submit option `spark.mesos.containerizer=mesos` but runs Dispatcher with default config (`False`, `True`)
- a user is experienced and enables UCR in both Dispatcher and jobs (`True`, `True`) 

The whole goal of this exhaustive testing is to make sure we don't miss some not-so-obvious combination of containerizers in which user propagation breaks.

tests/test_spark_users.py
akirillov: I created a Jira ticket for it, probably it makes sense to wait until we have proper dependency management for the testing package from [dcos-commons](https://github.com/mesosphere/dcos-commons).

PR #497, opened 2019-03-25
[DCOS-47213]: Spark: fix "jars" bug in submit_builder.go [spark-build]

farhan5900:

Resolves [DCOS-47213](https://jira.mesosphere.com/browse/DCOS-47213)

It was tested by running **make test**.


cli/dcos-spark/submit_builder_test.go
```diff
@@ -98,6 +98,14 @@ func (suite *CliTestSuite) TestProcessJarsFlag() {
 	actual, _ := transformSubmitArgs(inputArgs, args.boolVals)
 	assert.Equal(suite.T(), expected, actual)
 }
+
+func (suite *CliTestSuite) TestProcessJarsFlagIfJarsWordInAppPath() {
+	inputArgs := []string{"--conf spark.cores.max=8", "app/jars/main.jar", "100"}
+	expected := []string{"--conf spark.cores.max=8", "app/jars/main.jar", "100"}
```
akirillov: According to the original issue:
```
dcos spark run --submit-args="<some confs> http://s3.com/jars/app.jar" 

will be converted to:

--conf spark.mesos.uris=http://s3.com/jars/app.jar
--conf spark.driver.extraClassPath=/mnt/mesos/sandbox/app.jar
```
So we should expect the following correct behaviour:
```
inputArgs := "--conf spark.cores.max=8 --jars=http://s3.com/jars/dependency.jar main.jar 100"
expected := []string{"--conf=spark.cores.max=8",
	"--jars=http://s3.com/jars/dependency.jar",
	"--conf=spark.mesos.uris=http://s3.com/jars/dependency.jar",
	"--conf=spark.driver.extraClassPath=/mnt/mesos/sandbox/dependency.jar",
	"--conf=spark.executor.extraClassPath=/mnt/mesos/sandbox/dependency.jar",
	"main.jar"}
```

Please check `TestProcessJarsFlag` and similar tests

farhan5900: If you mention **--jars** flag as well as jars word in app jar path or somewhere else in **submit-args**, it will not break without the fix which I did. Because once the **jars** flag is checked to be present in `processJarsFlag` function then it is parsed and the loop is broken. 

It was breaking if we are not giving **jars** flag and word _jars_ appears somewhere else in the **submit-args** flag.

`TestProcessJarsFlag` and all similar tests are already doing the job of checking the **jars** flag correctness.

That is why, I have not included **jars** flag in the input.

akirillov: Yes, this is absolutely correct. What I'm saying is that the results of the call to this function are different, and more configuration parameters are created from the same input.

For example, in the source code of `func processJarFlags` these lines are [present](https://github.com/mesosphere/spark-build/blob/master/cli/dcos-spark/submit_builder.go#L372-L373):
```
newArgs = append(newArgs, "--conf", "spark.driver.extraClassPath=/mnt/mesos/sandbox/"+jarPaths)
newArgs = append(newArgs, "--conf", "spark.executor.extraClassPath=/mnt/mesos/sandbox/"+jarPaths)
```
It's not clear for me how after the call to this function we expect to receive this:
```
expected := []string{"--conf spark.cores.max=8", "app/jars/main.jar", "100"}
```

Looks like the equality assertion `assert.Equal(suite.T(), expected, actual)` doesn't work as expected too

farhan5900: If you see the function `processJarFlags`, there is a `IF` in it, which is checking whether **jars** flag exists or not. If **jars** flag is there, only then it will append all those extra configuration, otherwise it will return the same arguments it was pass to.

For example, if we pass argument as follows:

`inputArgs := []string{"--conf spark.cores.max=8", "app/jars/main.jar", "100"}`

There is no **jars** flag in the input, so it will return as it is:

`expected := []string{"--conf spark.cores.max=8", "app/jars/main.jar", "100"}`

but if you pass function argument having **jars** flag as follows:

`inputArgs := {"--conf spark.cores.max=8", "--jars http://s3.com/jars/dependecny.jar", "main.jar", "100"}`

It will return those extra configuration also:

```
expected := []string{"--conf=spark.cores.max=8",
	"--jars=http://s3.com/jars/dependency.jar",
	"--conf=spark.mesos.uris=http://s3.com/jars/dependency.jar",
	"--conf=spark.driver.extraClassPath=/mnt/mesos/sandbox/dependency.jar",
	"--conf=spark.executor.extraClassPath=/mnt/mesos/sandbox/dependency.jar",
	"main.jar"}
```

This is the current behavior of the `processJarsFlag` function.

akirillov: got it, thanks for the clarification. I'd suggest @samvantran  looking at this too

samvantran: Thanks for the ping and review @akirillov and thanks for taking up the work @farhan5900. You're both right in your arguments but the test we want to verify here is that only the `--jars` flag is processed and everything else is left alone (ie the main.app is ignored even if it contains the word `jar` in its path). So we can't only test the `processJars` method, we have to test `transformSubmitArgs` like Anton suggested.

This is because previously if the word `jars` was present in the path of the app, the CLI would barf (even if there was a `--jars` flag or not). What we want is if the word `jars` is present in both eg
```
inputArgs := {"--conf spark.cores.max=8", "--jars http://s3.com/jars/dependency.jar", "http://s3.com/jars/main.jar", "100"}
```
only `dependency.jar` is added with all the additional confs. We would get an `expected` like the following:
```
expected := []string{"--conf=spark.cores.max=8",
	"--jars=http://s3.com/jars/dependency.jar",
	"--conf=spark.mesos.uris=http://s3.com/jars/dependency.jar",
	"--conf=spark.driver.extraClassPath=/mnt/mesos/sandbox/dependency.jar",
	"--conf=spark.executor.extraClassPath=/mnt/mesos/sandbox/dependency.jar",
	"http://s3.com/jars/main.jar"}
```
You don't necessarily have to add a whole new test (though I argue you should to make the intent clear). You could modify one of the previous processJarsTests to have an app with `jars` in its path and verify that it isn't included in the added `confs`.

farhan5900: Thanks for the review @samvantran and @akirillov. I will modify one of the existing `TestProcessJarsFlag` test to check `jars` in app path.

PR #502, opened 2019-04-03
[DCOS-33733] Enable support of --packages flag in Spark CLI

farhan5900:

Resolves [DCOS-33733](https://jira.mesosphere.com/browse/DCOS-33733)

* Spark CLI will support `--packages` flag in `submit-args` flag of `run` sub-command.
* While processing the `--packages` flag, It automatically sets `spark.jars.ivy` property to `/mnt/mesos/sandbox/.ivy2` as default location to download the jars.
* Unit Tests to verify setting of these properties: `spark.jars.packages=<maven_coordinates>` and `spark.jars.ivy=<ivy_user_dir>`.
* Integration test to verify that the mentioned packages in the flag is being downloaded in driver and executor.
* Added a custom class `ProvidedPackages`, to support the integration test.

* Unit Tests covering all the scenario of giving value to the flag and setting `spark.jars.ivy` property in the process.
* Running integration test `test_packages_flag`

## Release Notes

* Added support of `--packages` flag in Spark CLI

### DIFF
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -220,7 +220,9 @@ Args:
 	val.flag(submit).StringVar(&val.s)
 	args.stringVals = append(args.stringVals, val)
 
-	val = newSparkVal("packages", "spark.jars.packages", "Comma-separated list of package names (i.e. maven coordinates) to be downloaded on the driver and executor classpaths.")
+	val = newSparkVal("packages", "spark.jars.packages", "Comma-separated list of maven coordinates of jars to include "+
+                "on the driver and executor classpaths. Will search the local maven repo, then maven central and any additional remote "+
+                "repositories given by --repositories. The format for the coordinates should be groupId:artifactId:version")
         val.flag(submit).StringVar(&val.s)
         args.stringVals = append(args.stringVals, val)
 
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -221,6 +221,10 @@ Args:
 	val.flag(submit).StringVar(&val.s)
 	args.stringVals = append(args.stringVals, val)
 
+	val = newSparkVal("packages", "spark.jars.packages", "Comma-separated list of colon-separated package names and versions to be downloaded on the driver and executor classpaths.")
+        val.flag(submit).StringVar(&val.s)
+        args.stringVals = append(args.stringVals, val)
+
 	val = newSparkVal("py-files", "spark.submit.pyFiles", "Add .py, .zip or .egg files to "+
 		"be distributed with your application. If you depend on multiple Python files we recommend packaging them "+
 		"into a .zip or .egg.")
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -220,7 +220,7 @@ Args:
 	val.flag(submit).StringVar(&val.s)
 	args.stringVals = append(args.stringVals, val)
 
-	val = newSparkVal("packages", "spark.jars.packages", "Comma-separated list of package names to be downloaded on the driver and executor classpaths.")
+	val = newSparkVal("packages", "spark.jars.packages", "Comma-separated list of package names (i.e. maven coordinates) to be downloaded on the driver and executor classpaths.")
         val.flag(submit).StringVar(&val.s)
         args.stringVals = append(args.stringVals, val)
 
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -305,6 +305,7 @@ func transformSubmitArgs(argsStr string, boolVals []*sparkVal) ([]string, []stri
 	sparkArgs, appArgs := make([]string, 0), make([]string, 0)
 
 	args = processJarsFlag(args)
+        args = processPackagesFlag(args)
 
 LOOP:
 	for i := 0; i < len(args); {
@@ -376,6 +377,30 @@ func processJarsFlag(args []string) []string {
 			jarPaths := strings.Join(jarNames, ":/mnt/mesos/sandbox/")
 			newArgs = append(newArgs, "--conf", "spark.driver.extraClassPath=/mnt/mesos/sandbox/"+jarPaths)
 			newArgs = append(newArgs, "--conf", "spark.executor.extraClassPath=/mnt/mesos/sandbox/"+jarPaths)
+			newArgs = append(newArgs, args[i+1:]...)
+			break
+		}
+	}
+	if len(newArgs) > 0 {
+		return newArgs
+	}
+	return args
+}
+
+// Special handling of '--packages' flag
+// Adds additional flag to set ivy user directory to instruct Spark to download jars to sandbox
+func processPackagesFlag(args []string) []string {
+	var newArgs []string
+	for i := 0; i < len(args); i++ {
+		if strings.HasPrefix(args[i], "--packages") {
+			if strings.Contains(args[i], "=") {
+				newArgs = append(newArgs, args[:i+1]...)
+			} else {
+				i++
+				newArgs = append(newArgs, args[:i+1]...)
+			}
+			// add --conf spark.jars.ivy=/mnt/mesos/sandbox/.ivy2
+			newArgs = append(newArgs, "--conf", "spark.jars.ivy=/mnt/mesos/sandbox/.ivy2")
 			newArgs = append(newArgs, args[i+1:]...)
 			break
 		}
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -22,6 +22,8 @@ var keyWhitespaceValPattern = regexp.MustCompile("(.+)\\s+(.+)")
 var backslashNewlinePattern = regexp.MustCompile("\\s*\\\\s*\\n\\s+")
 var collapseSpacesPattern = regexp.MustCompile(`[\s\p{Zs}]{2,}`)
 
+const mesosSandboxPath = "/mnt/mesos/sandbox"
+
 type sparkVal struct {
 	flagName string
 	propName string
@@ -374,9 +376,9 @@ func processJarsFlag(args []string) []string {
 			}
 			// add --conf spark.driver.extraClassPaths=one.jar:two.jar
 			// add --conf spark.executor.extraClassPath=one.jar:two.jar
-			jarPaths := strings.Join(jarNames, ":/mnt/mesos/sandbox/")
-			newArgs = append(newArgs, "--conf", "spark.driver.extraClassPath=/mnt/mesos/sandbox/"+jarPaths)
-			newArgs = append(newArgs, "--conf", "spark.executor.extraClassPath=/mnt/mesos/sandbox/"+jarPaths)
+			jarPaths := strings.Join(jarNames, ":"+mesosSandboxPath+"/")
+			newArgs = append(newArgs, "--conf", "spark.driver.extraClassPath="+mesosSandboxPath+"/"+jarPaths)
+			newArgs = append(newArgs, "--conf", "spark.executor.extraClassPath="+mesosSandboxPath+"/"+jarPaths)
 			newArgs = append(newArgs, args[i+1:]...)
 			break
 		}
@@ -399,8 +401,8 @@ func processPackagesFlag(args []string) []string {
 				i++
 				newArgs = append(newArgs, args[:i+1]...)
 			}
-			// add --conf spark.jars.ivy=/mnt/mesos/sandbox/.ivy2
-			newArgs = append(newArgs, "--conf", "spark.jars.ivy=/mnt/mesos/sandbox/.ivy2")
+			// add --conf spark.jars.ivy=path/to/ivy_user_dir
+			newArgs = append(newArgs, "--conf", "spark.jars.ivy="+mesosSandboxPath+"/.ivy2")
 			newArgs = append(newArgs, args[i+1:]...)
 			break
 		}
--- a/cli/dcos-spark/submit_builder.go
+++ b/cli/dcos-spark/submit_builder.go
@@ -119,7 +119,6 @@ appears to be client mode only? doesn't seem to be used in POST submit call at a
 - proxyUser:              --proxy-user SOMENAME
 client mode only (downloads jars to local system, wouldn't work for POST call):
 - ivyRepoPath:            (spark.jars.ivy)
-- packages:               --packages maven,coordinates,for,jars (spark.jars.packages)
 - packagesExclusions:     --exclude-packages groupId:artifactId,toExclude:fromClasspath (spark.jars.excludes)
 - repositories:           --repositories additional.remote,repositories.to.search
 yarn only (note: principal/tgt/keytab are supported in patched spark):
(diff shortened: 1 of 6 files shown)


cli/dcos-spark/submit_builder.go
```diff
@@ -221,6 +221,10 @@ Args:
 	val.flag(submit).StringVar(&val.s)
 	args.stringVals = append(args.stringVals, val)
```
samvantran: Also you probably want to delete this line: https://github.com/mesosphere/spark-build/blob/2157a6f86b6edd49c455b8d9dffdcc85d84f8c27/cli/dcos-spark/submit_builder.go#L122

farhan5900: I have deleted the line.

cli/dcos-spark/submit_builder_test.go
```diff
@@ -21,6 +21,7 @@ const keytabPrefixed = "__dcos_base64__keytab"
 const keytab = "keytab"
 const sparkAuthSecret = "spark-auth-secret"
 const marathonAppId = "spark-app"
+const sparkPackages = "group.one.id:artifact-one-id:version.one,group.two.id:artifact-two-id:version.two"
```
samvantran: Since this is only used in one test please move it inside the test method. It's better to encapsulate all the test data/logic closer together.

farhan5900: Moved it to the test method.

cli/dcos-spark/submit_builder.go
```diff
@@ -221,6 +220,10 @@ Args:
...
 
+	val = newSparkVal("packages", "spark.jars.packages", "Comma-separated list of package names to be downloaded on the driver and executor classpaths.")
```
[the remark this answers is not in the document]
farhan5900: changed it accordingly.

cli/dcos-spark/submit_builder_test.go
```diff
@@ -363,3 +364,38 @@ func (suite *CliTestSuite) TestSaslSecret() {
...
+func (suite *CliTestSuite) TestPackagesFlag() {
+	inputArgs := fmt.Sprintf(
+		"--packages %s "+
+		"--class %s "+
+		"%s --input1 value1 --input2 value2", sparkPackages, mainClass, appJar)
+
+	cmd := createCommand(inputArgs, image)
+        payload, err := buildSubmitJson(&cmd, marathonConfig)
+
+        m := make(map[string]interface{})
```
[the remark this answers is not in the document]
farhan5900: given a meaningful name

cli/dcos-spark/submit_builder_test.go
```diff
@@ -363,3 +364,38 @@ func (suite *CliTestSuite) TestSaslSecret() {
...
+                suite.T().Errorf("%s", err.Error())
+        }
+
+        stringProps := map[string]string{
+                "spark.mesos.task.labels":                    fmt.Sprintf("DCOS_SPACE:%s", marathonAppId),
+                "spark.ssl.noCertVerification":               "true",
+                "spark.executor.memory":                      "1G", // default
+                "spark.submit.deployMode":                    "cluster",
+                "spark.mesos.driver.labels":                  fmt.Sprintf("DCOS_SPACE:%s", marathonAppId),
+		"spark.jars.packages":                        sparkPackages,
```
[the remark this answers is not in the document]
farhan5900: we don't need to create whole JSON request only spark.jars.packages will also work.

tests/test_spark.py
```diff
@@ -175,6 +175,18 @@ def test_jars_flag(service_name=utils.SPARK_SERVICE_NAME):
...
+def test_packages_flag(service_name=utils.SPARK_SERVICE_NAME):
+    utils.run_tests(
+        app_url=utils.dcos_test_jar_url(),
+        app_args="20",
+        expected_output="The sum of 20 numbers is 210",
+        service_name=service_name,
+        args=["--conf spark.mesos.containerizer=mesos",
+              "--conf spark.jars.ivy=/mnt/mesos/sandbox/.ivy2",
+              "--packages com.google.guava:guava:23.0",
+              "--class ProvidedPackages"])
```
samvantran: shouldn't we add `nobody` as part of this test since you found an issue with that?

tests/jobs/scala/build.sbt
```diff
@@ -16,7 +16,8 @@ lazy val root = (project in file("."))
       "com.github.scopt" %% "scopt" % "3.7.0",
       "com.datastax.spark" %% "spark-cassandra-connector" % "2.0.7",
       "com.airbnb" % "kafka-statsd-metrics2" % "0.5.2+issue28" from "https://infinity-artifacts.s3.amazonaws.com/scale-tests/kafka-statsd-metrics2-0.5.2+issue28-20180612-8c1e3c4f3fa83.jar",
-      "com.datadoghq" % "java-dogstatsd-client" % "2.5"
+      "com.datadoghq" % "java-dogstatsd-client" % "2.5",
```
akirillov: do we really use it anywhere?

tests/jobs/scala/src/main/scala/ProvidedPackages.scala
```diff
@@ -0,0 +1,42 @@
...
+/**
+  * Application that outputs the sum of numbers from 1 to N.
+  * Specifically we want to test whether a package (e.g. guava) is available by specifying in the packages flag in submit-args.
+  * E.g. 
+  * dcos spark run --submit-args="\
+        --packages=com.google.guava:guava:23.0 \
+        --class ProvidedPackages \
+        <jar> number"
+  */ 
+object ProvidedPackages {
```
akirillov: please, add this application to a README with a short description of its purpose. It's good to maintain an index of all test apps to ease the navigation

tests/jobs/scala/src/main/scala/ProvidedPackages.scala
```diff
@@ -0,0 +1,42 @@
...
+	(1 to inputNumber).foreach(driverSum += _)
+
+        val spark = SparkSession.builder.appName(AppName).getOrCreate()
+
+	// Calculate the sum on the executor
+        val executorSum = spark.sparkContext.parallelize(1 to inputNumber).reduce((x, y) => IntMath.checkedAdd(x, y))
+        
+        // Verify the both sum
+        Verify.verify(driverSum == executorSum, "Sum of %s numbers on the executor (%s) is not equal to sum on the driver (%s)", inputNumber.toString, executorSum.toString, driverSum.toString)
+	println(s"The sum of $inputNumber numbers is $driverSum")
```
akirillov: let's avoid using custom text which is then checked outside and just output a single number. This way we can rely on the client only (python test) and avoid changing it when surrounding text changed.

tests/test_spark.py
```diff
@@ -175,6 +175,18 @@ def test_jars_flag(service_name=utils.SPARK_SERVICE_NAME):
...
               "--class MultiConfs"])
 
+@pytest.mark.sanity
+def test_packages_flag(service_name=utils.SPARK_SERVICE_NAME):
+    utils.run_tests(
+        app_url=utils.dcos_test_jar_url(),
+        app_args="20",
+        expected_output="The sum of 20 numbers is 210",
+        service_name=service_name,
+        args=["--conf spark.mesos.containerizer=mesos",
```
akirillov: Let's no override containerizer because for the time being the default one is Docker

tests/test_spark.py
```diff
@@ -175,6 +175,18 @@ def test_jars_flag(service_name=utils.SPARK_SERVICE_NAME):
         args=["--jars {}".format(uploadedJarUrl),
               "--class MultiConfs"])
 
+@pytest.mark.sanity
+def test_packages_flag(service_name=utils.SPARK_SERVICE_NAME):
+    utils.run_tests(
+        app_url=utils.dcos_test_jar_url(),
+        app_args="20",
+        expected_output="The sum of 20 numbers is 210",
```
akirillov: Let's change this to a number-only check.

tests/test_spark.py
```diff
@@ -175,6 +175,18 @@ def test_jars_flag(service_name=utils.SPARK_SERVICE_NAME):
...
+        service_name=service_name,
+        args=["--conf spark.mesos.containerizer=mesos",
+              "--conf spark.jars.ivy=/mnt/mesos/sandbox/.ivy2",
```
akirillov: This way users have to provide `--conf spark.jars.ivy` every time they use `--packages` with the CLI. The target goal is to specify it add this property in `dcos spark cli` when `--packages` is specified. Let's use `$MESOS_SANDBOX` env var instead of hardcoding as well.

cli/dcos-spark/submit_builder.go
```diff
@@ -221,6 +220,12 @@ Args:
 	val.flag(submit).StringVar(&val.s)
 	args.stringVals = append(args.stringVals, val)
 
+	val = newSparkVal("packages", "spark.jars.packages", "Comma-separated list of maven coordinates of jars to include "+
+                "on the driver and executor classpaths. Will search the local maven repo, then maven central and any additional remote "+
+                "repositories given by --repositories. The format for the coordinates should be groupId:artifactId:version")
+        val.flag(submit).StringVar(&val.s)
+        args.stringVals = append(args.stringVals, val)
```
akirillov: we need to add `--conf spark.jars.ivy` so that users don't worry about it

tests/jobs/scala/src/main/scala/ProvidedPackages.scala
```diff
@@ -0,0 +1,42 @@
...
+  * Specifically we want to test whether a package (e.g. guava) is available by specifying in the packages flag in submit-args.
+  * E.g. 
+  * dcos spark run --submit-args="\
+        --packages=com.google.guava:guava:23.0 \
+        --class ProvidedPackages \
+        <jar> number"
+  */ 
+object ProvidedPackages {
+    def main(args: Array[String]): Unit = {
+        val AppName = "ProvidedPackages App"
```
akirillov: Let's keep the name the same as class name. And, please make the val name start with lower-case because it's not a class-level constant (according to scalastyle).

cli/dcos-spark/submit_builder_test.go
```diff
@@ -363,3 +363,33 @@ func (suite *CliTestSuite) TestSaslSecret() {
...
+        jsonMap := make(map[string]interface{})
+
+        json.Unmarshal([]byte(payload), &jsonMap)
+
+        if err != nil {
+                suite.T().Errorf("%s", err.Error())
+        }
+
+        stringProps := map[string]string{
+		"spark.jars.packages": sparkPackages,
```
akirillov: when added, we need to check for `--conf spark.jars.ivy` property presence too

tests/jobs/scala/build.sbt
farhan5900: I don't know. That was already there.

tests/test_spark.py
farhan5900: I have added this one because job was failing with access denied issue.

tests/jobs/scala/src/main/scala/ProvidedPackages.scala
farhan5900: Removed the surrounding text and only printing final sum

tests/test_spark.py
farhan5900: Changed it to the number only check.

tests/test_spark.py
farhan5900: No need to specify `spark.jars.ivy` every time when using `--packages` flag. It will be added automatically.

cli/dcos-spark/submit_builder.go
farhan5900: Added `--conf spark.jars.ivy` while processing `--packages` flag in the CLI.

tests/jobs/scala/src/main/scala/ProvidedPackages.scala
farhan5900: Corrected the variable name

cli/dcos-spark/submit_builder_test.go
farhan5900: Added the required check

tests/jobs/scala/src/main/scala/ProvidedPackages.scala
farhan5900: Added short description, usage and example of the application.

tests/jobs/scala/build.sbt
akirillov: not critical for this PR but it would be nice to check if it used anywhere

tests/jobs/scala/README.md
```diff
@@ -72,3 +72,25 @@ In the example above Spark Driver will launch 4 single-core executors (controlle
 properties). When all executors are up, 12000 unique keys will be generated 4 times i.e. each of 4 partitions will contain 
 the same set of keys which will then be grouped by 4 reducers. `sleepBeforeShutdown` parameter is needed in order to
 verify tasks' properties while they're running and avoid quick application termination. 
+
+## ProvidedPackages
```
akirillov: 👍

tests/jobs/scala/build.sbt
```diff
@@ -16,7 +16,7 @@ lazy val root = (project in file("."))
       "com.github.scopt" %% "scopt" % "3.7.0",
       "com.datastax.spark" %% "spark-cassandra-connector" % "2.0.7",
       "com.airbnb" % "kafka-statsd-metrics2" % "0.5.2+issue28" from "https://infinity-artifacts.s3.amazonaws.com/scale-tests/kafka-statsd-metrics2-0.5.2+issue28-20180612-8c1e3c4f3fa83.jar",
-      "com.datadoghq" % "java-dogstatsd-client" % "2.5"
```
akirillov: we can't just drop the dependency without knowing if it used anywhere. The original comment was mostly about checking where it is used.

tests/test_spark.py
```diff
@@ -175,6 +175,16 @@ def test_jars_flag(service_name=utils.SPARK_SERVICE_NAME):
         args=["--jars {}".format(uploadedJarUrl),
               "--class MultiConfs"])
 
+@pytest.mark.sanity
+def test_packages_flag(service_name=utils.SPARK_SERVICE_NAME):
```
akirillov: `service_name` already defaults to `SPARK_SERVICE_NAME` in `utils.run_tests` no need to provide it here

cli/dcos-spark/submit_builder_test.go
```diff
@@ -21,6 +21,7 @@ const keytabPrefixed = "__dcos_base64__keytab"
 const keytab = "keytab"
 const sparkAuthSecret = "spark-auth-secret"
 const marathonAppId = "spark-app"
+const mesosSandbox = "/mnt/mesos/sandbox"
```
akirillov: `/mnt/mesos/sandbox` is local to a container and can potentially vary. Can we use `$MESOS_SANDBOX` environment variable which is set by executor to avoid binding to an absolute path?

tests/jobs/scala/build.sbt
farhan5900: I have checked and it seems it was not used anywhere, at-least directly. Do you wan me to add it back?

cli/dcos-spark/submit_builder_test.go
farhan5900: Since our CLI (`submit_builder.go`) is not using `$MESOS_SANDBOX` and hard coding that value, we have to do the same for the test also.
Will `$MESOS_SANDBOX` be available everywhere, where CLI is running?

tests/jobs/scala/build.sbt
farhan5900: Added it back.

tests/test_spark.py
farhan5900: Removed it.

tests/jobs/scala/build.sbt
farhan5900: Currently not used anywhere.

akirillov:
Thanks, @farhan5900! The changes look very solid. A few final things:
- can you please update the PR description to reflect how the changes have been tested
- include release notes (please check merged PRs for inspiration)
- finally, make sure CI builds are green

PR #505, opened 2019-04-16
[DCOS-33745] fetched cli segfault fix from dcos-commons, added a check for 401 response code

rpalaznik:

Resolves [DCOS-33745] - Unhelpful error message in CLI when not logged in (https://jira.mesosphere.com/browse/DCOS-33745)

- Updated vendored dependency on dcos-commons which includes a fix for segfault on empty responses (https://jira.mesosphere.com/browse/DCOS-42966 - Kubernetes CLI can segfault)
- Added a custom check for "401 Unauthorized" response, to print a more helpful error message 
`<HTTP Method> request to <URL> was not authorized. Make sure you are logged in.`
instead of 
`dcos-beta-spark: error: invalid character '<' looking for beginning of value, try --help`

- Manually tested
- Added a unit test for CLI change

## Release Notes
- [Bugfix] Improved error message in CLI when not logged into a cluster

### DIFF
--- a/cli/dcos-spark/main_test.go
+++ b/cli/dcos-spark/main_test.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"testing"
+
+	"github.com/mesosphere/dcos-commons/cli/client"
+	"github.com/stretchr/testify/assert"
+)
+
+func createHttpResponseWithStatus(statusCode int, filename string) http.Response {
+	status := fmt.Sprintf("%v %s", statusCode, http.StatusText(statusCode))
+	var body io.ReadCloser
+	if filename != "" {
+		content, _ := ioutil.ReadFile(filename)
+		body = ioutil.NopCloser(bytes.NewBuffer(content))
+	}
+	return http.Response{StatusCode: statusCode, Status: status, Body: body}
+}
+
+func TestUnauthorizedResponseCustomCheck(t *testing.T) {
+	response := createHttpResponseWithStatus(http.StatusUnauthorized, "testdata/responses/unauthorized.html")
+	testMethod := "GET"
+	testUrl := "https://example.com/marathon/v2/apps/spark"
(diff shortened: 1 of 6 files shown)


cli/dcos-spark/main_test.go
```diff
@@ -0,0 +1,33 @@
...
+	return http.Response{StatusCode: statusCode, Status: status, Body: body}
+}
+
+func TestUnauthorizedResponseCustomCheck(t *testing.T) {
+	response := createHttpResponseWithStatus(http.StatusUnauthorized, "testdata/responses/unauthorized.html")
+	response.Request = client.CreateServiceHTTPRequest("GET", "https://example.com/marathon/v2/apps/spark")
+
+	client.SetCustomResponseCheck(checkForUnauthorized)
+	_, err := client.CheckHTTPResponse(&response, nil)
+	assert.Contains(t, err.Error(), "Make sure you are logged in.")
```
akirillov: As we're using `response.Request.Method` and `response.Request.URL` in the error message it makes sense to assert the exact match instead of substring presence.

cli/dcos-spark/vendor/github.com/mesosphere/dcos-commons/cli/README.md
```diff
@@ -36,12 +36,12 @@ cd sdk/cli/
 To run the compiled CLI (the example below uses Linux, pick the correct binary for your platform):
 
 ```bash
-./dcos-service-cli-linux hello-world -h
+./dcos-service-cli-linux helloworld -h
```
akirillov: is it coming from the updated dependencies? we should think about including those as a git module to avoid this kind of diffs in PRs but this is out of the scope of this PR.

akirillov:
@rpalaznik, can you please update the PR description to be on par with other merged PRs? It is also missing the 'Release Notes' section which comes very handy when we prepare release docs.

cli/dcos-spark/main_test.go
rpalaznik: Ok, changed to the exact assertion.

PR #506, opened 2019-04-17
[DCOS-41589][Integration Tests] --supervise causing frameworkId conflicts in mesos cluster mode

farhan5900:

Resolves [DCOS-41589](https://jira.mesosphere.com/browse/DCOS-41589)

Added an integration test to check the supervise mode of the spark application as well as it's conflicts of framework Id's, when job is restarting.

It was tested running manually on the local machine.


tests/test_spark.py
```diff
@@ -389,3 +389,67 @@ def test_task_stdout():
         assert stdout_file["size"] > 0, "stdout file should have content"
     finally:
         sdk_install.uninstall(utils.SPARK_PACKAGE_NAME, service_name)
+
+
+@pytest.mark.sanity
+def test_supervise_conflict_frameworkid():
```
samvantran: overall, the test code looks good but teamcity CI fails because it looks you're trying to submit a spark job after the dispatcher has already been uninstalled [in the previous test](https://github.com/mesosphere/spark-build/pull/506/files#diff-2caedadc3f4e69445462b60b70423fcfR391). Suggest either reusing a previous dispatcher (saves time) or redeploying one for your test

tests/test_spark.py
```diff
@@ -389,3 +389,67 @@ def test_task_stdout():
...
+        svc = shakedown.get_service(job_service_name)
+        if present:
+            return svc is not None
+        else:
+            return svc is None
+
+    job_args = ["--supervise",
+                "--class", "MockTaskRunner",
+                "--conf", "spark.cores.max=8",
+                "--conf", "spark.executors.cores=4"]
```
akirillov: let's be more conservative and use 1 executor with 1 core. When we move to parallel test execution it will save other tests from resource starvation.

tests/test_spark.py
```diff
@@ -389,3 +389,67 @@ def test_task_stdout():
...
+
+    driver_id = utils.submit_job(app_url=utils.dcos_test_jar_url(),
```
akirillov: after submitting a driver, we need to wrap the following code in try-finally and kill the Driver when the test is over or failed in the middle. It'll save us from orphaned jobs in the cluster.

This pattern occurs more often so we'll address it separately and extract this handling into utils as a separate PR.

tests/test_spark.py
```diff
@@ -389,3 +389,67 @@ def test_task_stdout():
...
+                                 app_args="4 1800",
+                                 service_name=utils.SPARK_SERVICE_NAME,
+                                 args=job_args)
+    LOGGER.info("Started supervised driver {}".format(driver_id))
+    wait_job_present(True)
+    LOGGER.info("Job has registered")
+    sdk_tasks.check_running(job_service_name, 1)
+    LOGGER.info("Job has running executors")
```
akirillov: can we please rename `LOGGER` into `log` like in all newly developed tests

tests/test_spark.py
```diff
@@ -389,3 +389,67 @@ def test_task_stdout():
...
+
+    service_info = shakedown.get_service(job_service_name).dict()
+    driver_regex = "spark.mesos.driver.frameworkId={}".format(service_info['id'])
+
+    status, stdout = shakedown.run_command_on_agent(service_info['hostname'],
```
akirillov: while this is a valid approach, it has a drawback of multiple ssh trips to the agent host. I'd suggest using task exec instead, e.g.
```
dcos task exec <submission ID> "ps aux | grep java | <clear the noise with something like 'cut d' ' -f1'> | xargs kill -9
```

tests/test_spark.py
```diff
@@ -389,3 +389,67 @@ def test_task_stdout():
...
+        status, stdout = shakedown.run_command_on_agent(service_info['hostname'],
+                                                        "sudo kill -9 {}".format(pid),
+                                                        username=sdk_cmd.LINUX_USER)
+
+        if status:
+            print("Killed pid: {}".format(pid))
+        else:
+            print("Unable to killed pid: {}".format(pid))
+
+    wait_job_present(True)
```
akirillov: before that, we need to verify that the original task has been stopped/killed

tests/test_spark.py
```diff
@@ -389,3 +389,67 @@ def test_task_stdout():
...
+    LOGGER.info("Job has re-registered")
+    sdk_tasks.check_running(job_service_name, 1)
+    LOGGER.info("Job has re-started")
+
+    restarted_service_info = shakedown.get_service(job_service_name).dict()
+    assert service_info['id'] != restarted_service_info['id'], "Job has restarted with same framework Id"
+
+    kill_info = utils.kill_driver(driver_id, utils.SPARK_SERVICE_NAME)
```
akirillov: great that we check for valid termination as well 👍

tests/test_spark.py
farhan5900: As far as I understood, in this module dispatcher is auto installed, it will be reused. The teamcity CI fails because of [this](https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=1800891&_focus=19929) reason.

tests/test_spark.py
farhan5900: Reduced it to 1.

tests/test_spark.py
farhan5900: Added try-finally block.

tests/test_spark.py
farhan5900: Renamed `LOGGER` to `log`.

tests/test_spark.py
farhan5900: Added statement to check whether job is stopped/killed.

tests/test_spark.py
farhan5900: Thanks :+1:

tests/test_spark.py
farhan5900: `dcos task exec` is not working in this case it is saying `Command is allowed only for UCR`.
Instead of doing multiple ssh, changed it to make single ssh and kill the process.

tests/test_spark.py
samvantran: I saw [this error](https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=1800891&_focus=18660) which caused the diag bundle collection due to failure.

tests/test_spark.py
```diff
@@ -389,3 +389,64 @@ def test_task_stdout():
...
+
+        wait_job_present(True)
+        log.info("Job has registered")
+
+        sdk_tasks.check_running(job_service_name, 1)
+        log.info("Job has running executors")
+
+        service_info = shakedown.get_service(job_service_name).dict()
+        driver_regex = "spark.mesos.driver.frameworkId={}".format(service_info['id'])
+        status, stdout = shakedown.run_command_on_agent(service_info['hostname'],
```
akirillov: now it looks way cleaner, thanks. Keep in mind that this call causes SSH to an agent and given the known issue with `shakedown` might cause the sporadic failure of this or some irrelevant tests. I'd highly recommend switching to `dcos task exec` instead.

Also, we have a similar method implemented in `sdk_cmd.kill_task_with_pattern` - let's use that one instead

tests/test_spark.py
```diff
@@ -389,3 +389,64 @@ def test_task_stdout():
...
+        log.info("Started supervised driver {}".format(driver_id))
+
+        wait_job_present(True)
+        log.info("Job has registered")
+
+        sdk_tasks.check_running(job_service_name, 1)
+        log.info("Job has running executors")
+
+        service_info = shakedown.get_service(job_service_name).dict()
+        driver_regex = "spark.mesos.driver.frameworkId={}".format(service_info['id'])
```
akirillov: is `spark.mesos.driver.frameworkId=` part of regexp necessary?

tests/test_spark.py
```diff
@@ -389,3 +389,64 @@ def test_task_stdout():
...
+        log.info("Job has re-registered")
+        sdk_tasks.check_running(job_service_name, 1)
+        log.info("Job has re-started")
+
+        restarted_service_info = shakedown.get_service(job_service_name).dict()
+        assert service_info['id'] != restarted_service_info['id'], "Job has restarted with same framework Id"
+    finally:
+        kill_info = utils.kill_driver(driver_id, utils.SPARK_SERVICE_NAME)
+        log.info("{}".format(kill_info))
+        kill_info = json.loads(kill_info)
```
akirillov: let's avoid reassignment of the variable to a value with a different type. we can do `loads` in the assert statement the line below

tests/test_spark.py
```diff
@@ -389,3 +389,64 @@ def test_task_stdout():
...
+
+@pytest.mark.sanity
+def test_supervise_conflict_frameworkid():
+    job_service_name = "MockTaskRunner"
+
+    @retrying.retry(
+        wait_fixed=1000,
+        stop_max_delay=600 * 1000,
+        retry_on_result=lambda res: not res)
+    def wait_job_present(present):
```
akirillov: It looks like we keep reimplementing different flavours of this wait for each test, we should collect and extract all these waiting functions into a separate module. (not a part of this PR)

tests/test_spark.py
```diff
@@ -389,3 +389,64 @@ def test_task_stdout():
...
+        service_info = shakedown.get_service(job_service_name).dict()
+        driver_regex = "spark.mesos.driver.frameworkId={}".format(service_info['id'])
+        status, stdout = shakedown.run_command_on_agent(service_info['hostname'], 
+                "ps axo pid,command | grep -v grep | grep '{}' | cut -d' ' -f1 | sudo xargs kill -9".format(driver_regex),
+                username=sdk_cmd.LINUX_USER)
+
+        if status:
+            log.info("Killed all the PIDs")
+        else:
+            log.info("Unable to kill PIDs")
```
akirillov: let's remove this conditional logging after switching to `sdk_cmd.kill_task_with_pattern`

tests/test_spark.py
farhan5900: Simple `java` will not work, because one of the process is `docker run`. We can use `job_service_name` or only `frameworkId`.

tests/test_spark.py
farhan5900: changed `shakedown.run_command_on_agent` to `sdk_cmd.kill_task_with_pattern`.

tests/test_spark.py
farhan5900: moved it to the `assert` statement.

tests/test_spark.py
farhan5900: Removed it.

tests/test_spark.py
akirillov: @farhan5900, it seems that something is odd with the way the fixture works as pointed out by @samvantran. I'd suggest to do some debugging of the test itself and look at the cluster state at the moment of calling `utils.submit_job` by pausing the execution.

akirillov:
@farhan5900, it turned out that the behaviour of foldered Spark tests led to environment modifications which prevented supervised tests (and a few others if executed multiple times on the same cluster) fail. TL;DR - those tests uninstalled CLI and not always cleaned up ZK. This PR addresses these issues: https://github.com/mesosphere/spark-build/pull/514, please rebase on master after it is merged.

PR #507, opened 2019-04-19
[DCOS-41592] - Adds an integration test for Mesos labels support

rpalaznik:
[DCOS-41592] - Adds an integration test for Mesos labels support in Spark

Resolves [DCOS-41592] - [Integration Tests] Add Mesos labels support to the Spark Dispatcher (https://jira.mesosphere.com/browse/DCOS-41592)

Adds an integration test that 
- Submits a job with parameter `--conf spark.mesos.driver.labels=foo:bar`
- Examines `dcos task --json <driver task id>` output and asserts that the label is present

The new integration test was run locally targeting a dc/os cluster in AWS.

## Release Notes

n/a

### DIFF
--- a/tests/test_spark.py
+++ b/tests/test_spark.py
@@ -398,3 +398,16 @@ def test_task_stdout():
         assert stdout_file["size"] > 0, "stdout file should have content"
     finally:
         sdk_install.uninstall(utils.SPARK_PACKAGE_NAME, service_name)
+
+
+@pytest.mark.sanity
+def test_mesos_label_support():
+    driver_task_id = utils.submit_job(app_url=utils.SPARK_EXAMPLES,
+                                      app_args="150",
+                                      args=["--conf spark.cores.max=1",
+                                            "--conf spark.mesos.driver.labels=foo:bar", # pass a test label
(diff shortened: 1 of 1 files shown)


tests/test_spark.py
```diff
@@ -59,6 +59,19 @@ def test_task_not_lost():
...
+@pytest.mark.sanity
+def test_mesos_label_support():
+    driver_task_id = utils.submit_job(app_url=utils.SPARK_EXAMPLES,
+                                      app_args="150",
+                                      args=["--conf spark.cores.max=1",
+                                            "--conf spark.mesos.driver.labels=foo:bar", # pass a test label
+                                            "--class org.apache.spark.examples.SparkPi"])
+
+    driver_task_info = sdk_cmd._get_task_info(driver_task_id)
+    test_label_values = [label['value'] for label in driver_task_info['labels'] if label['key'] == 'foo']
```
akirillov: it would be great to make it a bit more readable. Is it possible to make it look something like that?
```
assert {'foo':'bar'} in driver_task_info['labels']
```

rpalaznik: changed as suggested

PR #516, opened 2019-05-15
DCOS-52812 Replace Oracle Java with OpenJDK for Spark

VladimirKravtsov:

Oracle server JRE replaced with OpenJDK8 JRE.

1. Spark Unit Tests are passed,
2. Spark Build Integration Tests DC/OS 1.13 Permissive are passed,
3. Spark packages were successfully assembled on OpenJDK8.

## Release Notes

build and runtime environments are migrated to OpenJDK 8


porridge:
I will let @akirillov review this, but I think we are leaning towards hosting a copy of the tarball on `downloads.m.c` @samvantran ?

samvantran:
Correct -- we're redistributing from our own site

VladimirKravtsov:
> This looks fine but there are failing tests? Is it related?

Only one test is failing -
https://teamcity.mesosphere.io/viewLog.html?buildId=1899293&tab=buildResultsDiv&buildTypeId=DataServices_Spark_PR_SparkBuild_113_strict#testNameId167945310167305385

And i heard that is known issue, which is randomly reproduces. I have restarted the build

VladimirKravtsov:
> Unfortunately, this change is not that straightforward and needs additional research. First, Spark distribution is built using Oracle JDK which brings surprises when running on OpenJRE. So this change requires updating all the build instructions for Spark distribution, Dockerfiles required for the build, and running Spark unit tests as a bare minimum to proceed. Also, moving forward, we'll need to support later versions of JDK which brings [another set of issues](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#jdk-11-compatibility-notes).
> 
> So I suggest putting this PR on pause for the time being or even closing it in favour of changes related to Scala 2.12 + JDK 11 updates.
> 
> @VladimirKravtsov, as a side-note, there's a reason for having PR description template - please follow it in the future.

I agree that Spark-build and other components should be tested and mb soak[ed] after migration on OpenJDK JRE. But I'm not agree that we are definitely run in some issues, because Oracle HotSpot JVM is based on OpenJDK HotSpot project. So, they are mostly the same VM, except that Oracle JVM has several additional commercial features like Java Flight Recorder, Application Class Data Sharing and Cooperative Memory Management. Which are not used by Scala. 

And I definitely don't speak about moving to JDK11 now, because it's still unsupported by Scala (as you mentioned).
Regarding to PR template - noted, but i have been creating the PR from IDE and the template hasn't been fetched for some reason.

akirillov:
Thanks for such a detailed explanation @VladimirKravtsov. Not that I'm questioning official documentation but we used to run into issues previously with the code compiled with Oracle **JDK** and executed on Open **JRE** - probably due to some system calls.

The good news is that the upstream Spark distribution is built with OpenJDK. So in order to move forward with Java 8 only, we need to rebuild our Spark distribution with it, run Spark unit tests, upload tested builds to `downloads.mesosphere` and re-run the integration tests from the current repo. I don't think there will be any showstoppers but a new distribution (not a release) is needed. Please follow _The Ultimate Spark Development Guide_ from the internal wiki to build it.

VladimirKravtsov:
@akirillov I have ran unit tests for spark on OracleJDK and OpenJDK and got the same results, but... on both runtimes it fails, please, take a look on the logs - 
[spark-ut-2.4.0-2.9.2-openjdk.log](https://github.com/mesosphere/spark-build/files/3205240/spark-ut-2.4.0-2.9.2-openjdk.log)
[spark-ut-2.4.0-2.9.2-oraclejdk.log](https://github.com/mesosphere/spark-build/files/3205241/spark-ut-2.4.0-2.9.2-oraclejdk.log)

akirillov:
Hi @VladimirKravtsov, based on the build logs, the problematic test was this one:
```
- SPARK-3697: ignore files that cannot be read. *** FAILED ***
  2 was not equal to 1 (FsHistoryProviderSuite.scala:177)
```
The problem was a bit unobvious, but looking at [this line in Spark codebase](https://github.com/mesosphere/spark/blob/custom-branch-2.4.x/core/src/test/scala/org/apache/spark/deploy/history/FsHistoryProviderSuite.scala#L174) it turned out that the test was failing when executed by `root` user which still could access the file after read permissions have been reset. 

Attached is a build log for Oracle JDK. Wiki description is updated and now shows how to run tests as a non-root user.

[spark-build-log-hadoop-2.9-oracle-jdk.tar.gz](https://github.com/mesosphere/spark-build/files/3210132/spark-build-log-hadoop-2.9-oracle-jdk.tar.gz)

VladimirKravtsov:
@akirillov Finally, unit tests  are passed successfully on OpenJDK 8 + hotspot. I run it over `custom-branch-2.4.x` branch for both - hadoop 2.7 and 2.9.
[spark-ut-2.4.0-2.7-openjdk-success.log](https://github.com/mesosphere/spark-build/files/3218586/spark-ut-2.4.0-2.7-openjdk-success.log)
[spark-ut-2.4.0-2.9-openjdk-success.log](https://github.com/mesosphere/spark-build/files/3218587/spark-ut-2.4.0-2.9-openjdk-success.log)

samvantran:
Changes look good @VladimirKravtsov but please follow the PR format for this project. There is a reason why we included a template. You can see an example in PR #520.

VladimirKravtsov:
> Changes look good @VladimirKravtsov but please follow the PR format for this project. There is a reason why we included a template. You can see an example in PR #520.

Anton already told to me that in his review, but I created the PR from IDE, and the template didn't appeared there. I noted that for future.

samvantran:
Thanks @VladimirKravtsov. It might be easier if you don't create PRs from your IDE or if you do then please edit the PR to include the proper description (including this one). You can copy it from other PRs.

VladimirKravtsov:
@akirillov I have uploaded both Spark packages (with hadoop 2.7 and 2.9 support) into downloads.mesosphere.io bucket and adjusted uris in `manifest.jso`, appropriately. Please, review it.

manifest.json
```diff
@@ -2,16 +2,16 @@
     "spark_version": "2.4.0",
     "default_spark_dist": {
         "hadoop_version": "2.9",
-        "uri": "https://downloads.mesosphere.com/spark/assets/2.8.0/spark-2.4.0-bin-hadoop2.9.2.tgz"
+        "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.4.0-bin-2.9.2-openjdk8-SNAPSHOT-20190530.tgz"
```
samvantran: These URIs should not include `SNAPSHOT`

VladimirKravtsov: Anton told to me add the SNAPSHOT postfix. We remove it eventually once the packages will be moved to production bucket. But Let me know if there is a reason get rid SNAPSHOT now.

samvantran: Yup, sorry about that. Feel free to disregard!

VladimirKravtsov: no problem :-)

akirillov: Please, re-upload artefacts to follow the pattern. `hadoop` suffix should be present to make the numbers clear. Tarball name should have a form `spark-2.4.0-bin-hadoop<hadoop version>-openjdk8-SNAPSHOT-<YYYYmmDD>`

VladimirKravtsov: ok, I fixed this, but such names were generated during the packages assembling. So, to avoid such naming issues in future, perhaps, we can adjust packages name generation algorithm, which uses during Spark packages building, to comply with this pattern?

akirillov: This is actually a good question. We're adding `hadoop` suffix to make it more readable for the customers using Spark. However, it is different from the upstream distribution which is more widely used. Changing the final name in the code doesn't seem right, because it will require upstreaming it which will change the standard naming for all the Spark users, rolling it back will mean that we'll be cherry-picking that commit forever. I think it's better to address it with build tooling.

VladimirKravtsov: Yes, once we will have appropriate CI Job for build Spark package, we can rename the archive with convenient format.

PR #519, opened 2019-05-20
[DCOS-47994] Flaky tests - Spark: test_checkpointing.test_structured_streaming_recovery

alembiewski:

Resolves [DCOS-47994](https://jira.mesosphere.com/browse/DCOS-47994)

According to our observations, some amount of messages were lost during publishing to Kafka topic. It could happen because of different reasons (whether broker is not ready, topic is not created or some network issues). This behavior causes flakiness in test's results.
The changes proposed in this request are aimed to improve `KerberizedKafkaProducer` in terms of message delivery. Changes also contain improved diagnostic information, which could be used later on for troubleshooting.

The flaky integration test were executed locally.


tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
```diff
@@ -35,14 +35,28 @@ object KerberizedKafkaProducer {
...
+      override def onCompletion(metadata: RecordMetadata, exception: Exception): Unit = {
+        if (exception != null) {
+          exception.printStackTrace()
+        } else {
+          println(s"Message has been published to (topic_partition@offset): ${metadata.toString}")
+        }
       }
+      })
     }
+    Thread.sleep(3000) // wait a few seconds for metrics update
```
akirillov: let's avoid using `Thread.sleep` and use scheduled executor service for metrics polling and terminating it when metrics appear.

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
```diff
@@ -35,14 +35,28 @@ object KerberizedKafkaProducer {
...
+    println("===========================================================================")
+    println("Producer metrics by topic:")
+    producer.metrics.values
+      .filter(m => m.metricName().group().equalsIgnoreCase("producer-topic-metrics"))
+      .foreach(metric => println(metric.metricName().description() + ": " + metric.metricValue()))
+    println("===========================================================================")
+    println(s"Partition metadata for topic '$topic':")
```
akirillov: can you please refactor the file and use `Logger` for logging instead of `println`?

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
```diff
@@ -35,14 +35,28 @@ object KerberizedKafkaProducer {
...
 
-      Try(producer.send(msg)) match {
-        case Success(_) => println("Message sent")
-        case Failure(ex) => ex.printStackTrace()
+      producer.send(msg, new Callback {
+      override def onCompletion(metadata: RecordMetadata, exception: Exception): Unit = {
+        if (exception != null) {
+          exception.printStackTrace()
+        } else {
+          println(s"Message has been published to (topic_partition@offset): ${metadata.toString}")
```
akirillov: is the `metadata` always present when there's no exception?

akirillov:
soft ping @alembiewski, let's avoid dangling PRs.

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
alembiewski: Yes, it can be `null` only when an error is occurred

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
alembiewski: sure. do you mean all the statements or only the new ones? It seems like it's no-brainer, making object extend `Logger` did the trick, but statements logged with `logInfo()` appeared in stderr log and I guess won't be displayed in test's output. Is it so?

akirillov: good catch! in pytests we grep the output from stdout, let's keep `println` then

alembiewski:
> does the information from metadata and partitions per topic provide us with the knowledge of how many messages have been received by Kafka? If not, it should be added for even better visibility.

`producer.metrics()` method returns only metrics, related to producer itself, like number of sent/retried messages. `producer.partitionsFor(topic)` shows information about each partition's metadata (leader, isr, replicas, offlineReplicas) for specified topic, but doesn't show information about number of received messages. 
I guess, we can count the number of recorded messages in callback method, because according to the source code comments: 
> @param callback A user-supplied callback to execute when the record has been **acknowledged by the server**

So, it is a guarantee, that message has been persisted on a server.
Message count can also be checked with `dcos kafka topic offsets streaming_test`. This command will show a number of messages for each partition:
`[
  {
    "0": "100"
  }
]`
I guess, we can use both options to verify message count.

akirillov:
thanks for the explanation, @alembiewski!  Let's add a simple counter to increment it in the callback and display the total number of acknowledged by the server messages. In order to use `dcos kafka` command from the test the fixture should install Kafka CLI alongside with the package: `dcos packa install kafka --cli --yes`

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
alembiewski: oh, this one is a bit tricky. when I troubleshooted the issue with producing to kafka, I observed, that the method `producer.metrics()`, which is invoked right after `send()` method, sometimes returns intermediate results, which means metric has already been added, but has not been updated in the moment of method invocation. `sleep` for a few seconds seems to solve this issue,  but yeah, I need to fine more elegant solution for that

alembiewski:
> Let's add a simple counter to increment it in the callback and display the total number of acknowledged by the server messages. 

done

tests/jobs/scala/src/main/scala/KerberizedKafkaProducer.scala
alembiewski: It seems like the issue with metrics has been resolved after switching to synchronous send method. I propose to use synchronous method in order to prevent various concurrency issues, that might impact test results as well.

akirillov:
@alembiewski, the switch to blocking `get` on the `Future` returned by producer `send` method might not solve the problem due to timing as well (are the metrics updated synchronously as well?). Another point is that developers rarely use blocking on `send` method (otherwise we wouldn't discover the issue). So I'd suggest implementing polling of producer metrics with a timeout and sane termination condition. We can terminate `KerberizedKafkaProducer` after e.g. 1 minute of polling in the case if producer metrics didn't show up or don't meet expectations.

PR #520, opened 2019-05-24
[DCOS-54419] Switched to reporting counters as gauges, unit tests refactoring

akirillov:

Resolves [DCOS-54419: Counters reported by Spark keep incrementing in Prometheus without actual value change](https://jira.mesosphere.com/browse/DCOS-54419)

As discovered during the dashboards review session, metrics having `counter` type kept incrementing their values in charts while the actual metric value hasn't changed in the code. It turned out that StatsD and Prometheus treat counter metrics differently (as increments in StatsD and as an actual value in Prometheus).

This PR changes counters reported type to `gauge` and introduces some update/refactorings to the unit tests.

* unit tests
* by verifying the fix in Prometheus

Before:
image

After:
image

## Release Notes
N/A


spark-statsd-reporter/src/main/java/org/apache/spark/metrics/sink/statsd/InstanceDetailsProvider.java
```diff
@@ -18,23 +18,29 @@
...
-                        new InstanceDetails(
-                                sparkConf.getAppId(),
-                                sparkConf.get("spark.app.name"),
-                                InstanceType.valueOf(SparkEnv.get().metricsSystem().instance().toUpperCase()),
-                                sparkConf.get("spark.executor.id"),
-                                sparkConf.get("spark.metrics.namespace", "default")
-                        );
+                instance = buildInstanceDetails();
             }
+            initialized = true;
```
samvantran: if the instance is null the first time, it's okay that it remain null for its lifetime? You never want to `buildInstanceDetails()` later ? I guess that would mean `SparkEnv` is never changed?

spark-statsd-reporter/src/main/java/org/apache/spark/metrics/sink/statsd/MetricType.java
```diff
@@ -1,11 +1,11 @@
 package org.apache.spark.metrics.sink.statsd;
 
 /**
+ * A subset of StatsD metric types used for Spark
  * @see <a href="https://github.com/etsy/statsd/blob/master/docs/metric_types.md">
  *        StatsD metric types</a>
  */
 interface MetricType {
-    String COUNTER = "c";
```
samvantran: There isn't any use for `COUNTERS` at all?

eg something like `totalJobsCompleted` for a dispatcher?

spark-statsd-reporter/src/main/java/org/apache/spark/metrics/sink/statsd/InstanceDetailsProvider.java
akirillov: This extra-check was introduced for lazy evaluation purposes and to avoid noisy logging every time it is `null`. The method returns `Optional.ofNullable(instance)` to indicate a possible absence of `SparkEnv` so it's fine for the instance to stay `null` (this use case is Dispatcher-specific, Driver and Executors always have `SparkEnv` initialized).

spark-statsd-reporter/src/main/java/org/apache/spark/metrics/sink/statsd/MetricType.java
akirillov: There's a difference between codahale/dropwizard counter metric type and StatsD counter. Codahale counters used a lot across the codebase and mutated programmatically. However, when you try to report a counter type to StatsD, it treats it as an increment only and always increases target metric value by adding the value of codahale/dropwizard counter.

So in order to report counter values, they need to become gauges to avoid StatsD increments. This [comment](https://github.com/mesosphere/spark-build/blob/DCOS-54419-switch-statsd-to-gauges-and-tests-refactoring/spark-statsd-reporter/src/main/java/org/apache/spark/metrics/sink/statsd/StatsdReporter.java#L108-L109) contains similar explanation and an example

spark-statsd-reporter/src/main/java/org/apache/spark/metrics/sink/statsd/InstanceDetailsProvider.java
samvantran: ah thanks for the explanation!

spark-statsd-reporter/src/main/java/org/apache/spark/metrics/sink/statsd/MetricType.java
samvantran: > However, when you try to report a counter type to StatsD, it treats it as an increment only and always increases target metric value by adding the value of 

This makes sense to me. Counters by nature increment. Gauges report values that can increase/decrease. My question was more along the lines of "do we never report counter types to StatsD"? Or maybe codahale counter metrics are being mutated incorrectly when they should be treated/renamed as gauges.

Anyway, it's not an important point to debate; just a question on removing `COUNTER` as a type.

PR #529, opened 2019-06-07
DCOS-54708: UCR containerizer by default

rpalaznik:

Resolves [DCOS-54708](https://jira.mesosphere.com/browse/DCOS-54708)

* Changed default value for `UCR_containerizer` parameter to `true`

* By running existing integration tests

## Release Notes

* Enabled UCR containerizer by default for Spark Dispatcher

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -54,13 +54,13 @@ dcos package install spark --cli
 # Custom Installation
 
 You can customize the default configuration properties by creating a JSON options file and passing it to `dcos package
-install --options`. For example, to launch the Dispatcher using the Universal Container Runtime (UCR), create a file
-called `options.json`:
+install --options`. For example, to launch the Dispatcher using Docker instead of the Universal Container Runtime (UCR),
+create a file called `options.json`:
 
 ```json
 {
   "service": {
-    "UCR_containerizer": true
+    "UCR_containerizer": false
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -54,13 +54,13 @@ dcos package install spark --cli
...
-install --options`. For example, to launch the Dispatcher using the Universal Container Runtime (UCR), create a file
-called `options.json`:
+install --options`. For example, to launch the Dispatcher using Docker instead of the Universal Container Runtime (UCR),
+create a file called `options.json`:
 
 ```json
 {
   "service": {
-    "UCR_containerizer": true
+    "UCR_containerizer": false
```
samvantran: I think we should take this out. The readme here is about providing `custom installation` and using a simple example of changing the cpu or memory would suffice. IMO if we're now defaulting to UCR, it is confusing to provide examples where we switch it off (thereby turning off metrics, opening the door for docker issues, etc)

akirillov: @samvantran, agreed. We should start removing mentions of Docker containerizer from the docs and default to UCR. Also, IIRC we have a ticket for removing `docs` folder from here and providing a link to DCOS Service docs to avoid doing double work syncing two sources (they've already diverged significantly). I suggest keeping this change for the sake of docs integrity in the scope of the current PR.

samvantran: Here you go: [DCOS-37659](https://jira.mesosphere.com/browse/DCOS-37659). Not to pollute the current sprint plans, I vote to add the doc ticket to the next sprint and let @rpalaznik continue with his current sprint work. Should he deliver everything early, then we can pull in the docs work knowing that all other commitments were met. WDYT?

akirillov: @samvantran, SGTM!

PR #536, opened 2019-08-21
[DCOS-54709] Scala 2.12 support

rpalaznik:

Resolves [DCOS-54709](https://jira.mesosphere.com/browse/DCOS-54709) - Add Scala 2.12 support for Spark

- Updated Makefile, and bin/jenkins-*.sh scripts to support multiple scala versions
- Updated manifest.json with new spark distributions with scala 2.12
- Default spark variant set to use hadoop-2.9 and scala 2.12
- Added a smoke test for custom spark docker images

I wanted to avoid creating test tags in git, so I tested the changes manually without using any production credentials, except uploading new spark distributions to downloads.mesosphere.io domain.

1)  Run the following script to build spark distribution archives:
```
SPARK_VERSION=2.4.3 \
DIST_DESTINATION_DIR=/tmp \
bin/jenkins-dist-publish.sh
```
Executed in `spark-tools` container, as described in the wiki. I had `mesosphere/spark` and `mesosphere/spark-build` repositories checked out and mounted next to each other. I've added DIST_DESTINATION_DIR variable to skip upload to S3 and save the archives locally.

2) Uploaded the spark distributions to downloads.mesosphere.io and updated manifest.json

3) Run the following script to build and push docker images to my personal docker hub account
```
DOCKER_USERNAME=rpalaznik \
DOCKER_PASSWORD=[redacted :)] \
SPARK_BUILD_VERSION=2.4.3 \
DOCKER_DIST_IMAGE=rpalaznik/spark \
GITHUB_DISABLE=true \
bin/jenkins-package-publish.sh
```

As I mentioned, this is just to test that the script is working. When making an actual release, you should simply create a release tag and run the jenkins job as described in the wiki.

4) Run the smoke test for custom docker images using the instructions from the wiki on how to run integration tests and by passing these variables
- `TEST_SH_SMOKE_TEST_SPARK_DOCKER_IMAGES='rpalaznik/spark:2.4.3-scala-2.11-hadoop-2.7,rpalaznik/spark:2.4.3-scala-2.11-hadoop-2.9,rpalaznik/spark:2.4.3-scala-2.12-hadoop-2.7,rpalaznik/spark:2.4.3-scala-2.12-hadoop-2.9'`
- `PYTEST_ARGS='-k test_spark_docker_images'`

If `TEST_SH_SMOKE_TEST_SPARK_DOCKER_IMAGES` is not set, e.g. during a CI integration test run, the smoke test will be skipped.

## Release Notes

- Added support for Spark distributions built with Scala 2.12
- Default spark distribution is built with Scala 2.12 and Hadoop 2.9 profiles

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3,28 +3,28 @@
     "default_spark_dist": {
         "scala_version": "2.12",
         "hadoop_version": "2.9",
-        "uri": "https://rpalaznik-dev.s3.amazonaws.com/dcos-54709/spark-2.4.0-bin-scala-2.12-hadoop-2.9.tgz"
+        "uri": "https://downloads.mesosphere.io/spark/assets/spark-2.4.3-bin-scala-2.12-hadoop-2.9.tgz"
     },
     "spark_dist": [
         {
             "scala_version": "2.12",
             "hadoop_version": "2.9",
-            "uri": "https://rpalaznik-dev.s3.amazonaws.com/dcos-54709/spark-2.4.0-bin-scala-2.12-hadoop-2.9.tgz"
+            "uri": "https://downloads.mesosphere.io/spark/assets/spark-2.4.3-bin-scala-2.12-hadoop-2.9.tgz"
         },
         {
             "scala_version": "2.12",
             "hadoop_version": "2.7",
-            "uri": "https://rpalaznik-dev.s3.amazonaws.com/dcos-54709/spark-2.4.0-bin-scala-2.12-hadoop-2.7.tgz"
+            "uri": "https://downloads.mesosphere.io/spark/assets/spark-2.4.3-bin-scala-2.12-hadoop-2.7.tgz"
         },
         {
             "scala_version": "2.11",
             "hadoop_version": "2.9",
-            "uri": "https://rpalaznik-dev.s3.amazonaws.com/dcos-54709/spark-2.4.0-bin-scala-2.11-hadoop-2.9.tgz"
+            "uri": "https://downloads.mesosphere.io/spark/assets/spark-2.4.3-bin-scala-2.11-hadoop-2.9.tgz"
         },
         {
             "scala_version": "2.11",
             "hadoop_version": "2.7",
-            "uri": "https://rpalaznik-dev.s3.amazonaws.com/dcos-54709/spark-2.4.0-bin-scala-2.11-hadoop-2.7.tgz"
+            "uri": "https://downloads.mesosphere.io/spark/assets/spark-2.4.3-bin-scala-2.11-hadoop-2.7.tgz"
         }
     ]
 }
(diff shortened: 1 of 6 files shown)


tests/jobs/python/shuffle_app.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import print_function
+import sys
+from random import getrandbits
+from time import sleep
+
+from pyspark.sql import SparkSession
+
+# ShuffleApp test job ported from Scala to Python
```
akirillov: While this job serves as a good smoke test - it doesn't verify Scala 2.12 support. Let's instead build two distributions of our tests jar: 2.11 and 2.12 and expose Scala version as a parameter so we can test Spark 2.12 with integration tests jar 2.12 and Spark 2.11 with 2.11

tests/test_spark_distributions.py
```diff
@@ -0,0 +1,49 @@
...
+
+log = logging.getLogger(__name__)
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+SMOKE_TEST_SPARK_DOCKER_IMAGES = os.getenv('SMOKE_TEST_SPARK_DOCKER_IMAGES', '')
+
+
+@pytest.mark.sanity
+@pytest.mark.smoke
+def test_spark_docker_images():
+    if not SMOKE_TEST_SPARK_DOCKER_IMAGES:
```
akirillov: 👍 in the future instead of publishing jars and including them in `resource.json` we'll publish different docker images so we can utilize this test for all of them

manifest.json
```diff
@@ -1,17 +1,30 @@
 {
     "spark_version": "2.4.0",
     "default_spark_dist": {
+        "scala_version": "2.11",
         "hadoop_version": "2.9",
-        "uri": "https://downloads.mesosphere.com/spark/assets/spark-2.4.0-bin-hadoop2.9.2-openjdk8-SNAPSHOT-20190530.tgz"
+        "uri": "https://downloads.mesosphere.io/spark/assets/spark-2.4.3-bin-scala-2.11-hadoop-2.9.tgz"
```
akirillov: A few comments on this:
- let's please use `mesosphere.com` instead of `.io`
- let's keep asset naming consistent, e.g.: `spark-2.4.3-bin-hadoop-2.9-scala-2.11-SNAPSHOT-20190822.tgz` (timestamp is important to have an idea of when this build has been created)
- let's avoid using prod assets bucket for non-release distributions and store artifacts in `infinity-artifacts` bucket

manifest.json
```diff
@@ -1,17 +1,30 @@
 {
     "spark_version": "2.4.0",
```
akirillov: shouldn't that be updated to `2.4.3` too?

manifest.json
rpalaznik: Updated the naming and moved to `infinity-artifacts` bucket

manifest.json
rpalaznik: Couldn't find where this property is used or visible. Updated anyway though.

PR #540, opened 2019-09-10
[DCOS-58390] Quota enforcement support for Spark Dispatcher

akirillov:

Resolves [DCOS-58390: Propagate role/quota configuration from environment variables to Dispatcher](https://jira.mesosphere.com/browse/DCOS-58390)

This PR adds a new service property `enforce_role` which instructs Spark Dispatcher to reject all submissions that attempt to override it. This PR comes together with role enforcement PR for Spark https://github.com/mesosphere/spark/pull/66.

The new role assignment behaviour now has the following logic:

- if `MESOS_ALLOCATION_ROLE` is set and `MARATHON_APP_ENFORCE_GROUP_ROLE=TRUE` then Spark Dispatcher will be launched with `spark.mesos.role=${MESOS_ALLOCATION_ROLE}` and `spark.mesos.dispatcher.role.enforce` set to `true` which will have the following semantics:
  - Spark Dispatcher registers with Mesos using `MESOS_ALLOCATION_ROLE`
  - if `spark-submit` doesn't provide `spark.mesos.role` the group role will be used for Drivers
  - if `spark-submit` provides `spark.mesos.role` which is different from the group role, the submission will be rejected
  - if `spark-submit` provides `spark.mesos.role` equal to the group role, the submission will be accepted

if the previous condition doesn't hold true (i.e. a group role is not enforced), Dispatcher will be launched with `spark.mesos.role=${SPARK_DISPATCHER_MESOS_ROLE}` which comes from `config.json`:
- if `SPARK_DISPATCHER_ENFORCE_ROLE` is set to `true` then Spark Dispatcher will reject all submissions that attempt to override the role it is running with itself
  - the semantics of role enforcement is the same as for group role, the only difference is the role itself 

- tests from this repo
- specific tests implemented for testing quota which were run against DCOS 1.14 which has quota functionality in place: [test_quota_permissive.log](https://github.com/mesosphere/spark-build/files/3598321/test_quota.log), [test_quota_strict.log](https://github.com/mesosphere/spark-build/files/3601627/test_quota_strict.log)

## Release Notes

- [New Feature] Added Spark Dispatcher role propagation and enforcement support
- [New Feature] Added Marathon group role enforcement support


akirillov:
@vinodkone, it would be really great if you can confirm the described behaviour of the Dispatcher with regards to role enforcement. Thanks in advance!

docker/runit/service/spark/run
```diff
@@ -76,6 +72,19 @@ function configure_properties() {
         add_if_non_empty spark.mesos.dispatcher.driverDefault.spark.executorEnv.VIRTUAL_NETWORK_ENABLED "${VIRTUAL_NETWORK_ENABLED}"
     fi
 
+    if [[ "${MESOS_ALLOCATION_ROLE}" != "" ]] && [[ "${MARATHON_APP_ENFORCE_GROUP_ROLE}" = "TRUE" ]]; then
```
samvantran: ~Where does `MARATHON_APP_ENFORCE_GROUP_ROLE` come from? Can't seem to find it anywhere else in the code~
nvm, found it in the Marathon code [here](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/mesos/TaskBuilder.scala)

akirillov:
Thanks, @mpereira. An integration test is WIP now, will update the PR once finished

akirillov:
Merged this PR as all tests in all PR builds are green

PR #544, opened 2019-09-16
Modify cluster setup method to fix failing 'hdfs-kerberos-delete-terasort-files' job on soak114

alembiewski:

Fixes failing job (https://soak114s.testing.mesosphe.re/#/jobs/detail/spark-hdfs-kerberos-terasort) on soak114 cluster. 

Before the main job execution, the clean-up job `hdfs-kerberos-delete-terasort-files` gets created, but it fails to execute with `no cluster attached` error. After some investigation, it was decided to replace  the old python script, used in soak110, with `dcos cli` command. 

Tested on majority of soak clusters with overriding `$GIT_BRANCH` job parameter. 

## Release Notes

n/a

### DIFF
--- a/tests/test_soak.py
+++ b/tests/test_soak.py
@@ -75,7 +75,7 @@ TERASORT_DELETE_JOB_KERBEROS = {
         "cmd": " && ".join([
             "/bin/bash",
             "cd $MESOS_SANDBOX",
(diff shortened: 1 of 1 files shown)


alembiewski:
@akirillov, yes, the script could be found in `soak-cluster-configurations` repository, and it looks like it was simply copy-pasted from one soak cluster folder to another without any modifications.

PR #543, opened 2019-09-13
[DCOS-58773] Updated libmesos to 1.14

akirillov:
## What changes were proposed in this pull request?

Resolves [DCOS-58773: Spark frameworks with CNI enabled failed to authenticate with the Mesos master](https://jira.mesosphere.com/browse/DCOS-58773)

This PR updates `libmesos` bundle version which comes with proper handling of TLS for DCOS 1.14 
and virtual networks.
## How were these changes tested?

- integration tests from this repo
- DCOS 1.14 tests run independently from the PR branch:
  - [permissive](https://teamcity.mesosphere.io/viewLog.html?buildId=2279011&tab=buildResultsDiv&buildTypeId=DataServices_Spark_Nightly_DCOS_master_SparkBuild_master_permissive)
  - [strict](https://teamcity.mesosphere.io/viewLog.html?buildId=2279012&tab=buildResultsDiv&buildTypeId=DataServices_Spark_Nightly_DCOS_master_SparkBuild_master_strict)

## Release Notes

[Update] Updated `libmesos` to 1.14
[Update] Updated SDK (dcos-commons) to 0.57.0

PR #546, opened 2019-09-25
[DCOS-29717] The TLS tests for the data services should use different…

mediapills:
Resolves [DCOS-29717: The TLS tests for the data services should use different values for service_accounts and service_account_secret](https://jira.mesosphere.com/browse/DCOS-29717)

This PR merge values for service_accounts and service_account_secret during tests.

- integration tests from this repo

## Release Notes
- [Update] Updated setup security functionality during tests


mediapills:
Related to [#3192](https://github.com/mesosphere/dcos-commons/pull/3192) (dcos-common)

akirillov:
Thanks for the PR @mediapills. Can you please update the PR title so it:
- follows the template (example: https://github.com/mesosphere/spark-build/pull/543)
- contains motivation for the change in its description

porridge:
I think we should just drop this PR and it will just migrate naturally during the next bump.

akirillov:
@porridge, unfortunately, this repo doesn't depend on `dcos-commons` as on a submodule but instead has a copy of `dcos-commons` testing code. This change looks like a port from `dcos-commons` which would be nice to have if this is the case.

porridge:
Right, most of our repos still have just a copy.
But don't you do periodic refreshes? If you only cherry pick selected changes it might become a problem at some point..

mediapills:
> Thanks for the PR @mediapills. Can you please update the PR title so it:
>     * follows the template (example: #543)
>     * contains motivation for the change in its description
Done, could you please verify?

akirillov:
Thanks, @mediapills. LGTM 👍 

@porridge, we have migration to dcos-commons submodule planned for the nearest future so cherry-picking won't be a problem anymore.

PR #556, opened 2019-10-16
[DCOS-59720] Introduce Spark Job for MWT

alembiewski:

Resolves [DCOS-59720 [DS] [Spark Operator] Create a better Spark Job for MWT](https://jira.mesosphere.com/browse/DCOS-59720)

This PR introduces two Spark applications:
1) `DatasetGenerator` - creates a dataset with specified record count and record size and writes the result on s3 bucket.
2) `DatasetSort` - reads data from s3 location and perform sort operation on the obtained `Dataframe`

- Spark Applications were tested locally;
- By running integration tests in CI 

## Release Notes

n/a


tests/jobs/scala/src/main/scala/DatasetGenerator.scala
```diff
@@ -0,0 +1,57 @@
+import org.apache.spark.sql.SparkSession
```
akirillov: let's create a separate `sorting` package for both generator and sorter with a README so they're not mixed up with the rest of the testing jobs

tests/jobs/scala/src/main/scala/DatasetGenerator.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import scala.util.Random
+
+/**
+	* Generates dataset in form of randomly generated key/value pairs. Can be used in stress-tests and benchmarks
+  *
+	* Usage: DatasetGenerator [totalRecords] [recordSize] [s3_destination]
```
akirillov: I'd suggest adding the following parameters:
- `numFiles` to control the resulting number of files created (maps to the number of partitions)
- `staticPrefixLength` to control the length of the static part of the keys 
- `randomSiffuxLength` to control the length of the variable part of the keys
- rename `recordSize` -> `valueSizeBytes`

akirillov:
Thanks, @alembiewski. Sorter and Gerenerator look good and were battle-tested during scale-tests. However, we have CI tests failing and it looks like the failures caused by `hadoop-aws` version bump which deprecates `s3n` URLs.  The culprit seems to be here: https://github.com/mesosphere/spark-build/blob/master/spark-testing/spark_s3.py#L9-L16.

PR #559, opened 2019-10-31
Benchmark updates to trigger shuffle and validate results

akirillov:

Resolves [DCOS-60317: Improve sorting benchmark to exercise shuffle](https://jira.mesosphere.com/browse/DCOS-60317).

This PR improves the existing sorting application in the following ways:
* using explicit schema to properly read binary field (it's been converted to a string by default - even after casting)
* switching to RDD API to avoid driver OOMs (potentially due to the fact that `ColumnStats` were adding significant overhead) and `Dataset`/`DataFrame` sorting results not being written in parallel by executors
* custom sorting `Ordering` leads to full shuffle (no optimizations involved as with `StringOrdering` when sorting by key)
* the results are written back to S3 to verify that the whole dataset is sorted and persisted in parallel by executors (which requires a shuffle operation)

This PR adds validation to verify:
* keys ordering within partitions (each persisted file should be sorted)
* global ordering (the order of keys is preserved across all persisted files) 

* by running TeraSort scale test from [mesosphere/kudo-spark-operator](https://github.com/mesosphere/kudo-spark-operator/tree/master/scale-tests) on 100GB input dataset and verifying it completes sorting and validation without errors

* validation log  [attached](https://github.com/mesosphere/spark-build/files/3816853/validation.log)

* performance graphs:

image

image

image

image


akirillov:
This PR is under active development at the moment. @alembiewski, can you please take a look at the changes?

A few things here:
```
spark
	.read
	.option("compression", "gzip")
	.schema(DataItem.schema)
	.json(inputPath)
	.as[DataItem]
	.rdd
	.sortBy(identity)(ord = ordering, ctag = ClassTag(classOf[DataItem]))
	.toDS()
	.write
	.json(outputPath)
```
- using explicit schema to properly read binary field (it's been converted to a string by default - even after casting)
- switching to RDD API to avoid driver OOMs (potentially due to `ColumnStats` being collected). With `Dataset`/`DataFrame` sorting results are not written in parallel by executors
- custom sorting `Ordering` leads to full shuffle (no optimizations involved as with `StringOrdering` when sorting by key)
- the results are written back to S3 to verify that the whole dataset is sorted and persisted

After adding the validation step, it turned out that not all the partitions are fully sorted. Some relevant information from [SPARK-19352](https://jira.apache.org/jira/browse/SPARK-19352):

>DataFrameWriter doesn’t allow users to write data out orderly, so unfortunately you can’t do this. This makes sense because sorted data files are not useful: when you read it back, Spark can’t get the ordering information back and optimize for it.

akirillov:
This PR is ready for review. The changes proposed in it were tested on 100GB dataset and 10 executors (1CPU, 8GB RAM). Sorting and validation completed successfully, see pictures attached to the description.

tests/jobs/scala/src/main/scala/sorting/DatasetGenerator.scala
```diff
@@ -10,6 +11,13 @@ import scala.util.Random
...
+object DataItem {
+	def schema = StructType(Seq(
+		StructField("key", StringType, nullable = false),
+		StructField("value", DataTypes.BinaryType, nullable = false)
+	))
+}
+
 case class Config(numFiles: Int = -1,
 									numRecords: Int = -1,
 									staticPrefixLength: Int = 120,
```
samvantran: Not really part of your PR but can we fix this wild formatting? Not sure why all of these params are so far away. (future idea: might be good to add a linter for the Scala code to avoid these issues)

tests/jobs/scala/src/main/scala/sorting/SortingApp.scala
```diff
@@ -1,18 +1,48 @@
...
 	*/
 
 object SortingApp {
 
+	val pattern: UnanchoredRegex = """"key":"([a-zA-Z0-9]+)"""".r.unanchored
+
+	case class PartitionValidationResult(
+																				fileName: String,
+																				isSuccess: Boolean,
+																				totalKeys: Long,
```
samvantran: Some more extreme whitespace

tests/jobs/scala/src/main/scala/sorting/SortingApp.scala
```diff
@@ -1,18 +1,48 @@
...
+																				outOfOrderKeys: Long,
+																				sample: List[String],
+																				firstKey: String,
+																				lastKey: String) extends Ordered[PartitionValidationResult] {
+
+
+		override def toString: String = {
+			s"PartitionValidationResult {\n\tfile: $fileName,\n\tsuccess: $isSuccess,\n\ttotalKeys: $totalKeys,\n\tfirst: $firstKey,\n\tlast: $lastKey,\n\tunorderedKeys: $outOfOrderKeys,\n\tsample:\n${sample.mkString("\n")}}"
```
samvantran: Maybe multi-line string here for readability? Eg
```
"""
PartitionValidationResult {
  file: $fileName,
  ...
"""
```

tests/jobs/scala/src/main/scala/sorting/SortingApp.scala
```diff
@@ -21,21 +51,88 @@ object SortingApp {
...
-			.sort("key")
-
-		spark.time(sortedDS.count())
+		val ordering = new Ordering[DataItem] {
+			override def compare(x: DataItem, y: DataItem): Int = x.key.compareTo(y.key)
+		}
+		//TODO: ideally, the time should be reported as a metric
+		spark.time {
+			spark
+				.read
```
samvantran: I'm starting to suspect you've got a golang formatter working in your IDE

tests/jobs/scala/src/main/scala/sorting/DatasetGenerator.scala
akirillov: wild formatting fixed here and everywhere

tests/jobs/scala/src/main/scala/sorting/SortingApp.scala
akirillov: good point, updated

akirillov:
Thanks for the review, @samvantran! I'll wait for builds to finish and will merge the PR ignoring 4 failing tests in DCOS 2.0 because they're not related to the change and the issue is being addressed in https://github.com/mesosphere/spark-build/pull/561

PR #560, opened 2019-11-01
Improved MockTaskRunner to produce heavier CPU load

akirillov:

Resolves [DCOS-60732: Improve MockTaskRunner to provide higher CPU utilization](https://jira.mesosphere.com/browse/DCOS-60732)

This PR adds a synthetic loop instead of `Thread.sleep`. This almost-empty loop provides high CPU usage and contains an artificial check to avoid any optimizations by JIT compiler.

* by running two versions of `MockTaskRunner` side-by-side and monitoring system+user time:
  
  MTR CPU usage


akirillov:
thanks for the great catch, @samvantran! It looks like all the role enforcement and foldered installation tests fail due to the role enforcement:
```
Bad request: Dispatcher is running with role enforcement enabled but submitted Driver attempts to override the default role. Enforced role: path, Driver role: *"
```

Nightlies are failing too, so I'll create a fix PR separately.

akirillov:
@samvantran, I'm about to merge this PR because these 4 failing tests in DCOS 2.0 builds are not related/caused by the change and that issue is being addressed separately in https://github.com/mesosphere/spark-build/pull/561

PR #561, opened 2019-11-06
[DCOS-60748] Spark Integration Test failing on DCOS 2.0 due to logic change for role and group

farhan5900:

Resolves [DCOS-60748](https://jira.mesosphere.com/browse/DCOS-60748)

It introduces changes in the existing integration tests specifically for the `Foldered Installation of Spark Job` and `Quota`. In DCOS 2.0, the group and role are enforced by default in case of foldered installation of the service. So the Dispatcher is running with the specific role enforced, not * as role. To submit the Spark Job, we have to use that role name as well.

In case of `Quota` test, this PR has just modified the scenarios to make test pass. For the proper fix we can have discussion about the approach that we should apply.

Ran `make test` on different versions of DCOS cluster.


tests/test_quota.py
```diff
@@ -90,8 +90,7 @@ def test_marathon_group_enforced(create_group, setup_spark, group, enforce_group
 
 @pytest.mark.sanity
 @pytest.mark.parametrize('role,enforce_role', [
-    ("spark_role", False),
-    ("custom_role", False)
+    ("dev", False)
```
akirillov: these use cases are still valid for DCOS versions prior to 2.0. Can we somehow refactor the test and mark it to be run with all roles only on DCOS < 2.0?

It looks like we can separate two use-cases:
- _any_ role propagation (DCOS pre-2.0)
- group role propagation (DCOS 2.0 and higher)

tests/test_quota.py
```diff
@@ -103,7 +102,7 @@ def test_dispatcher_role_propagated(setup_spark, role, enforce_role):
 
 @pytest.mark.sanity
 @pytest.mark.parametrize('role,enforce_role', [
-    ("spark_role", True)
```
akirillov: Custom role enforcement is a valid use case for earlier DCOS versions (pre-2.0). What if we refactor the tests and split them into two:
- _a_ role enforced (DCOS pre-2.0)
- group role enforced (DCOS 2.0 and higher)

Tests can be marked with `@pytest.mark.dcos_min_version('version')`

tests/test_quota.py
```diff
@@ -43,82 +41,93 @@ def create_group(group, enforce_group):
...
-@pytest.mark.parametrize('group,enforce_group,role,enforce_role', [
-    ("dev", True, "spark_role", False),
-    ("dev", True, "spark_role", True),
-    ("dev", True, "dev", False),
-    ("dev", True, "dev", True)
+@pytest.mark.parametrize('group,enforce_group,service_name,role,enforce_role', [
+    ("dev", True, "/dev/spark", "spark_role", False),
+    ("dev", True, "/dev/spark", "spark_role", True),
+    ("dev", True, "/dev/spark", "dev", False),
+    ("dev", True, "/dev/spark", "dev", True)
```
akirillov: 👍

PR #582, opened 2020-10-27
Stop agent's domain socket, which leads to systemd restarting the age…

alexeygorobets:
Fix for stopping Mesos agent in tests.

Once we take down the Mesos agent, if a task with an HTTP executor is running on the node, the executor will immediately attempt to connect to the agent's domain socket, which leads to systemd restarting the agent.

In order to prevent this, it looks like we now need to stop the Mesos agent using:
`sudo systemctl stop dcos-mesos-slave.service dcos-mesos-slave.socket`
Then, restarting the agent can be accomplished with:
`sudo systemctl start dcos-mesos-slave.service dcos-mesos-slave.socket`

Related ticket - [D2IQ-69935](https://jira.d2iq.com/browse/D2IQ-69935)


rishabh96b (approved):
LGTM

